PR #15, opened 2026-06-20
Add filter for uae metadata files in scan_tree

sonnenscheinchen:
This change skips adding useless *.uaem files created by Amiberry and FS-UAE into the directory hard drive. https://github.com/BlitterStudio/amiberry/wiki/Host-Directory-Filesystem-Metadata


LinuxJedi (approved):
Great idea. Many thanks!

PR #39, opened 2026-06-25
dirfs encoding fix

sonnenscheinchen:
This change tries to convert OS specific file system encoding to latin1 encoding used by the Amiga. Background: Currently Copperline just uses into_encoded_bytes() which returns whatever the host OS uses, UTF-8 in many cases. This results in incorrect file- and directory names containing non ASCII characters, eg. `portuguÃês` on the in-memory FFS volume.
I only tested this on Linux, but I don't see any reason why it shouldn't work on Mac or Windows.


LinuxJedi (approved):
Great find, many thanks.

PR #58, opened 2026-06-28
Do not require a ROM file if --load-state argument provided

sonnenscheinchen:
...because the ROM is included in the state file


LinuxJedi (approved):
100% Great catch :)

PR #73, opened 2026-06-30
Zipped floppy image support

sonnenscheinchen:
This change adds support for ADF-images in zip files. Many collections, like TOSEC, use that format, so it makes sense to support it. The implementation is quite simple: Only the local file header is parsed, not the central directory. Only the first file of the ZIP-archive is used, which is usually sufficient.


LinuxJedi (approved):
I love this idea. Many thanks!

PR #88, opened 2026-07-02
Reset Zorro boards to unconfigured state on warm reset

### LINKED ISSUE #87: [Bug]: Zorro boards vanish after first warm-reboot.
### Area

### Summary

Looks like Zorro boards don't go back to their initial configuration state after hitting ctrl-A-A, or after the Amiga spontaneously reboots.

### Copperline version

ad131bee

### Host system

Linux 7.2rc, AMD Ryzen AI

### Machine/configuration

 - Regular A1200
 - Kickstart 3.2
 - A590/A2091 SCSI controller
 - 2MB Fast RAM (is this Z2 RAM?)

### Reproduction steps

1. Hold both mouse buttons (or HELP / F11) to enter the Early Startup menu
2. Verify that boards appear in Expansion Board Diagnostic
3. Exit the early startup menu by either rebooting or booting
4. Now enter Early Startup again

and...

### Actual behaviour

Now the boards vanished!

### Expected behaviour

Zorro boards should reset to config state on cold / warm reset.

### Logs or debugger output

```text

```

codewiz:
Zorro boards vanished after a warm boot (Ctrl-Amiga-Amiga or a software reboot): the autoconfig chain stayed in the Configured state across reset, so Kickstart's autoconfig scan at $E80000 found no unconfigured board and rebuilt an empty ConfigDev list.

On real hardware both the keyboard reset and the CPU RESET instruction assert /RST on the expansion bus, which clears every board's autoconfig latch (including shut-up boards) while expansion RAM keeps its contents.

Add ZorroChain::warm_reset() that returns all boards to Unconfigured without zeroing RAM, and call it from both warm-reset paths. The RESET instruction path previously only reset COPCON and the floppies, so it also gains the ZorroDevice::reset loop the keyboard path already had.

Closes https://github.com/LinuxJedi/Copperline/issues/87

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -2355,6 +2355,13 @@ impl Bus {
     pub fn reset_custom_chips_from_cpu_reset(&mut self) {
         self.agnus.reset_copcon();
         self.floppy.reset_external_drives();
+        // RESET asserts /RST on the expansion bus: boards drop back to
+        // unconfigured (Kickstart re-runs autoconfig on the way back up)
+        // and expansion devices see a hardware reset.
+        for dev in &mut self.devices {
+            crate::zorro_device::ZorroDevice::reset(dev);
+        }
+        self.mem.zorro.warm_reset();
     }
 
     fn effective_diwhigh(&self) -> DiwHigh {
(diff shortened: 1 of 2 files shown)


LinuxJedi (approved):
Many thanks for finding and fixing this.

PR #111, opened 2026-07-04
disasm: decode MOVEC with control-register names

codewiz:
The UI disassembler showed MOVEC as DC.W $4E7A/$4E7B followed by a bogus decode of the extension word, which made CPU-detection code (exactly the code paths #91 dealt with) unnecessarily hard to read in the debugger.

Decodes both directions with control-register names across all models (SFC/DFC/CACR/TC, the TTRs, USP/VBR/CAAR/MSP/ISP, the 040 MMU registers, and the 060 BUSCR/PCR); unknown codes print as $NNN. Includes tests for both directions and the unknown-code case.

Split out of the debugger follow-up branch so it can merge independently; the model-aware address-mask work will follow as a separate PR.

### DIFF
--- a/src/disasm.rs
+++ b/src/disasm.rs
@@ -49,6 +49,32 @@ impl Stream<'_> {
     }
 }
 
+/// MOVEC control-register name for the extension word's Rc field, across
+/// all models (010 through 060); unknown codes print as $NNN.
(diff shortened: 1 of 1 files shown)


LinuxJedi (approved):
Great find. Many thanks!

PR #112, opened 2026-07-04
debugger: compare and display CPU addresses through the model's bus mask

codewiz:
UI_ADDR_MASK hardcoded A0-A23 across the interactive debugger, so on 020+ machines a breakpoint or watch set at a Zorro III address aliased its 24-bit shadow in chip RAM (both compare sides were masked, so $40001000 and $00001000 were the same breakpoint), and every surface that prints CPU addresses (double-fault OSD, watch/writer reports, stack traces, the memory panel base) truncated 32-bit addresses.

InteractiveBreaks now takes the owning machine's address-bus mask at construction (InteractiveBreaks::new, from address_mask_for_model), the GDB stub and window/console surfaces normalize through machine.ui_addr_mask(), and chip-DMA surfaces (copper breakpoints, blitter pointers) deliberately stay 24-bit. The FIND command still searches the 24-bit space as documented.

Regression test: a Z3 breakpoint under a 32-bit mask no longer matches its chip-RAM alias; the existing 24-bit toggle/mask test is unchanged.

Follow-up to #91's debugger fix; independent of #111.

### DIFF
--- a/src/cpu.rs
+++ b/src/cpu.rs
@@ -279,7 +279,7 @@ impl M68kMachine {
             dbg_crash_on: crate::envcfg::flag("COPPERLINE_DIAG_CRASH"),
             dbg_crash_dumped: false,
             dbg: crate::debugger::Debugger::from_env(),
-            ui_breaks: crate::debugger::InteractiveBreaks::default(),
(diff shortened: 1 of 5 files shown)


LinuxJedi (approved):
Completely forgot I needed to fix this. I hit it and then got distracted. Many thanks!

PR #113, opened 2026-07-04
debugger: peek device-board windows where side-effect-free

codewiz:
Bus::peek_word_any() routed RAM-backed Zorro regions but fell through device-backed windows, so every debugger surface (console TASKS/MEM, the GDB stub's memory reads) saw zeros or junk where the emulated CPU reads the A2091's boot ROM. Symptom: scsi.device sub-task and node names - ln_Name pointers into the board window's ROM half - displayed as 30 unprintable characters in TASKS.

Adds ZorroDevice::peek_word(), a side-effect-free peek defaulting to None: the A2091 answers for its ROM range (>= $2000) and keeps the DMAC/SBIC registers opaque, and peek_word_any() consults it for device windows. A2065/WASM boards keep the default.

Test: A2091 peeks return ROM bytes in the ROM half and None over registers.

### DIFF
--- a/src/a2091.rs
+++ b/src/a2091.rs
@@ -374,6 +374,16 @@ impl crate::zorro_device::ZorroDevice for A2091 {
         Self::write(self, off, size, value, host.memory_mut())
     }
 
+    fn peek_word(&self, off: u32) -> Option<u16> {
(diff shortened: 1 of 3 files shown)


LinuxJedi (approved):
Excellent! Many thanks!

PR #116, opened 2026-07-04
agnus: model sprite DMA at the register level

LinuxJedi:
## Summary

Sprite DMA is now modelled at the register level the way Agnus works,
replacing the descriptor-precompute model (vAmiga
executeFirstSpriteCycle/executeSecondSpriteCycle/updateSpriteDMA
semantics, ground-truthed against the real-A500 sprenacpu2 photos, which
match vAmiga exactly):

- Each channel keeps Agnus-side SPRxPOS/SPRxCTL register copies and
  vertical comparator values, updated identically by DMA fetches and
  CPU/Copper pokes (POS = vstart low bits; CTL = vstart high bit + whole
  vstop + disarm). Line-start comparators set/clear the per-channel DMA
  flip-flop -- vstop clears it **even when SPREN is off**, which is what
  leaves a sprite dead until the next field when software disables DMA
  across its vstop line (the sprena/sprdis mechanism).
- The vstop line's two slots fetch the next POS/CTL control words; the
  vertical-blank reset (PAL $19 / NTSC $14) forces every vstop to that
  line, which is how each field's first control-word fetch happens. No
  descriptor chain, no precomputed stream end: terminators, inverted
  vstop pairs, deferred starts and equal start/stop descriptors all fall
  out of the comparators naturally.
- SPRxPT advances only on words actually fetched (the skew model is
  gone; the frame-start frontier is just the live pointer), and an armed
  channel with DMA off keeps redisplaying its latches until a CTL
  fetch/poke disarms it -- this renders the vAmigaTS sprena/sprdis
  vertical-bar staircase that the old model structurally could not
  produce.
- Deleted with the descriptor model: register-stream seeding, SPRxPT
  retarget math, same-slot descriptor-restart discard, the save-state
  reconstruction hack (the register state is chip state and is now
  serialized/restored directly), ~250 net lines removed.
- Sprite bus-slot arbitration follows the same rules (enabled channel or
  vstop-line control fetch claims the slots, vblank-gated).
- `VideoPipelineStats` (host perf telemetry) is no longer serialized:
  call chunking legitimately differs across a save-state load.
  STATE_VERSION 17.

## vAmigaTS

Structural: sprenacpu2 now renders the yellow armed-persistence bar
staircase that the A500 photo and vAmiga show (the old model painted a
drifting streak instead). The family's mismatch **percentages stay
~30-45%**: those sweeps step CPU DMACON writes at 2-colour-clock
resolution over a CPU-timed flickering background, so the residual is
dominated by the known CPU-write-timing class (the #102/#104 residual),
not sprite modelling. interfere1/1b improved (7.2/5.3 -> 6.1/7.2 mixed),
manual1/2 and the collision buckets are unchanged.

## Testing

- Full suite green (1296 tests), clippy/fmt clean.
- Sprite tests rewritten to hardware flows: control words load at the
  vertical-blank reset line (shared helper), pointer reloads land before
  it, inverted vstop now pins the control-word refetch on the vstop line
  (vAmiga semantics) instead of the old run-to-bottom clamp.
- Byte-identity vs the two-slot-split baseline (#114): kick13boot 15s,
  Gen-X 110s, Hamazing, ITM, A1200 Kickstart, Zool -- identical.
  Second Nature 40s is the known benign animation-phase drift (the FSM
  build reaches the byte-identical frame 0.04s earlier).
- Sprite-sensitive regression demos vs current main: eon at 108s (the
  wolf/SPREN-clear scene class) byte-identical; Roots II AGA 60s
  byte-identical; Roots II ECS byte-identical at 20/35/50s with a
  timeline knife-edge crossing between 50s and 60s (same benign drift
  class as Second Nature).

### DIFF
--- a/docs/internals/chipset.md
+++ b/docs/internals/chipset.md
@@ -47,59 +47,47 @@ Alice adds the FMODE wide-fetch latch, which scales the bitplane and
 sprite fetch quanta (FMODE=0 stays byte-identical to the OCS/ECS slot
 timing).
 
+Sprite DMA is modelled at the register level, the way the chips work:
+there is no separate "descriptor" concept. Each channel keeps Agnus-side
+copies of its SPRxPOS/SPRxCTL words and the vertical comparator values
+derived from them, updated identically by DMA fetches and by CPU/Copper
+pokes (POS supplies the vertical start's low bits, CTL the high bit, the
+whole vertical stop, and disarms the display latches). At each line start
+the comparators run: passing vstart sets the channel's DMA flip-flop,
+passing vstop clears it -- even when SPREN is off, which is what leaves a
+sprite dead until the next field when software disables DMA across its
+vstop line. Writes landing on the matching line re-evaluate immediately.
+
 A sprite line's two DMA slots ($15+4N and $17+4N) are evaluated at their
-own colour clocks: the first slot runs the line's descriptor/arming logic
-and samples the DATA word(s) at that slot's beam time, the second samples
-DATB at its own beam time and assembles the display line, so chip RAM
-rewritten between the two slots is seen by DATB but not DATA. DMACON's
-SPREN is sampled by each slot individually, so a mid-line edge fetches
-exactly one word of the pair; the skipped word's slot leaves SPRxPT one
-fetch behind and the data stream shifts accordingly, while the display
-line reuses the stale latch on that side (a missed DATA slot never arms
-the sprite).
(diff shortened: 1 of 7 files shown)


LinuxJedi:
Note: this branch is stacked on #114 (the two-slot fetch split) and includes its commit; merging #114 first will reduce this diff to the FSM commit alone, or this PR can be merged on its own and #114 closed.

PR #121, opened 2026-07-05
serial: bidirectional TCP sink (mode = "tcp")

codewiz:
Adds a fourth [serial] mode bridging Paula's serial port to a host TCP listener, like UAE's TCP: serial device (same default port, 1234; override with `[serial] listen = "host:port"`).

With `Mount AUX:` + `NewShell AUX:` on the Amiga side, a connected client gets a full remote AmigaDOS console -- handy for scripted testing and for driving the emulated machine headlessly. One client at a time; a new connection replaces a finished one, and output with no client connected is dropped like an unplugged cable.

Implementation notes:
- A background thread owns the accept loop and the socket read half, staging input bytes through a channel with an atomic fill counter, so Paula's idle fast path (`has_pending_input`) stays a cheap atomic load and never touches a socket.
- Writes go directly to the client stream with TCP_NODELAY for shell interactivity.
- Zero new dependencies (std::net only), so it works on all platforms, unlike a pty-based bridge.
- Unit test covers the input/output round trip; tested interactively against an AUX: shell with minicom (via a pty bridge) and scripted clients.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -241,6 +241,10 @@ pub enum SerialMode {
     /// Serial in/out is bridged to host MIDI endpoints. Requires a build
     /// with the `midi` feature; without it, resolving this mode is an error.
     Midi,
+    /// Serial in/out is bridged to a host TCP port, like UAE's `TCP:`
+    /// serial device. With an `AUX:` shell on the Amiga side, a connected
+    /// client gets a remote AmigaDOS console.
+    Tcp,
 }
 
 impl SerialMode {
@@ -250,6 +254,7 @@ impl SerialMode {
             Self::Off => "off",
             Self::Stdout => "stdout",
             Self::Midi => "midi",
+            Self::Tcp => "tcp",
         }
     }
 }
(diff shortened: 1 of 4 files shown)


codewiz:
I have regrets about using TCP: I couldn't get a working Amiga shell with nc nor with screen, probably due to mismatches between VT100-ish terminal escapes and the ANSI escapes used by the shell.

Amiberry / WinUAE also relay the built-in serial to localhost:1234, which is probably a good choice for file transfers and modem emulation, but I think we should also have a pty mode that works directly with `minicom -D /tmp/amiga-serial`.

codewiz:
Thanks for the review - all four addressed in 582d7cf:

- **`buffered` unsigned underflow** (both comments): switched to `AtomicIsize`, so a `read_byte` that consumes a byte before the reader thread's matching `fetch_add` dips to a transient `-1` debt instead of wrapping to `usize::MAX` and getting stuck-true.
- **`nc` hint mangles IPv6**: now formats `local_addr.ip()`/`local_addr.port()` separately (and uses the resolved `local_addr`, so port-0 binds print the real port).
- **launcher drops `[serial] listen` on re-save**: `MachineSetup` now carries `serial_listen` through `from_raw`/`to_raw`; added a round-trip regression test.

LinuxJedi:
Sorry @codewiz, just merged a bunch of things and it caused a merge conflict. Can you sort that and then I should be able to get this in.

codewiz:
No problem - rebased onto current main (the conflict was just the new `[audio]` lines landing next to the `[serial]` block in `MachineSetup::to_raw`). Builds clean, fmt/clippy happy, serial tests pass. Ready when you are.

LinuxJedi (approved):
Excellent stuff!

PR #127, opened 2026-07-06
audio: Cmd/Alt+Shift+A cycles the live audio output

hobbo91:
Adds a keyboard shortcut for the runtime "Audio Out" menu item (from #124): `Cmd+Shift+A` on macOS, `Alt+Shift+A` on Linux/Windows, cycling the live output forward -- Default -> host devices -> Disabled -- the same as the menu, reusing `cycle_audio_output()`. It's guarded so it only fires on a running machine, not the config-screen placeholder or while a menu/panel is open.

Shift (rather than a bare `Cmd+A`) matches the scheme's tier for infrequent, set-and-forget actions (save/load state, warp limit), and avoids the "Select All" muscle-memory collision accidentally silencing or misrouting audio.

Documented in the in-window Keyboard Shortcuts panel, the UI guide table, and the README audio section.

Verification: clippy `-D warnings` + fmt clean (default and `--no-default-features`); 1319 tests, 0 failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,8 +189,8 @@ default output; `--list-audio-devices` prints the alternatives and
 `--audio-device NAME` (or `[audio] output_device`) selects one by
 case-insensitive substring, falling back to the default if it disappears. The
 device is also selectable in the configuration screen and switchable live from
-the in-window menu, which additionally offers "Disabled" to turn sound off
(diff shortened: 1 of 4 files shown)


LinuxJedi (approved):
Good idea!

Issue #155, opened 2026-07-09
[Bug]: 68040/68030 MMU access-fault continuation drops a store trapped by MMULib (MuForce/MuGA), corrupting data


codewiz:
## Area

Emulation accuracy (chipset/CPU behaviour)

## Summary

On the 68040 (and 68030), when MMULib's MuForce/MuGuardianAngel traps a
user-mode store to a page it has poisoned for monitoring, Copperline
discards the store instead of letting it complete. Data is silently
lost; with the write re-issued it can also hang. Real 68040 hardware and
WinUAE/Amiberry complete the store.

## Copperline version

`main` / v0.10.0 (`7965f81`); the m68k MMU core is unchanged on the
branch I tested (`feat/scsi-controller-select`, `d808f5a`).

## Host system

Linux (Fedora 45), x86-64.

## Machine/configuration (abbreviated)

```toml
[cpu]
model = "68040"   # also reproduces with model = "68030"
fpu = true
[machine]
profile = "A1200"
# AmigaOS 3.2 boot running MMULib: MuForce + MuGuardianAngel from S:Startup
# plus an A4091 (NCR 53C710) SCSI board used by the reproducer tool
```

## Reproduction steps

1. Boot AmigaOS 3.2 with MMULib active (`MuForce RAWIO AREGCHECK
   STACKCHECK`, then `MuGuardianAngel ... DUMPWALL`).
2. Run a program that performs user-mode stores into a buffer that
   overlaps a MuGA-poisoned page. My trigger is the `ncr7xx -t5` A4091
   diagnostic (DMA buffer allocated in fast RAM); it fails intermittently
   depending on whether the buffer lands on a poisoned page.

## Actual behaviour

Intermittent data corruption, e.g. the `DMA test` reports `Source data
changed from <pattern> to <stale>` -- the CPU's store never reached
memory. The trapped store is a user-mode long write (`SSW=0x0401`); the
monitored page reads back MuGA's poison `0xBADFEED0` (PDT=0).
Copperline's 68040 access-error RTE path, seeing the handler clear the
writeback-status valid bit, sets a one-shot suppression and drops the
restarted instruction's matching store (`mmu_write_suppress`,
`core/decode.rs`, format-7 frame). The 68030 path does the equivalent
via the SSW DF bit.

## Expected behaviour

The store should complete, as on real hardware and WinUAE/Amiberry
(which pass this exact MuForce+MuGA setup). WinUAE's
[...]

PR #156, opened 2026-07-09
m68k: report 68040 write faults in writeback slot 3

codewiz:
## What

The 68040 access-error stack frame builder pushed a faulted write into
writeback slot 2 (WB2S/WB2A/WB2D), and the RTE handler read the pending
writeback status back from that slot.

Real 68040 silicon and the WinUAE/Amiberry MMU core put a normal write
fault in slot **3**. See Amiberry `src/cpummu.cpp:434`:

```c
regs.wb3_status = write ? 0x80 | (ssw & 0x7f) : 0;
regs.wb3_data = val;
regs.wb2_status = 0;
```

WB2 is reserved for MOVE16 cacheline writes. Frame push order confirmed
at `src/newcpu_common.cpp:1442-1455` (WB3D at +$1C, WB3A at +$18, WB3S
at +$0E).

With the data in the wrong slot, a monitor that completes an allowed
store by copying WB3D to WB3A (MuGuardianAngel) reads a zero WB3D and
clobbers the target with 0.

## Change

- `exceptions.rs`: build the write into slot 3 (WB3S/WB3A/WB3D), clear WB2.
- `decode.rs`: the RTE absorb path reads WB3S/WB3A from +$0E/+$18.
- `access_fault_frame_tests.rs`: updated to assert the slot-3 layout (8 tests pass).

## Scope

This corrects the writeback slot to match the reference; it is **not** a
fix for the separate access-fault continuation issue tracked in #155, and
no functional repro isolates this change from that bug. It stands on the
frame unit tests.

### DIFF
--- a/crates/m68k/src/core/decode.rs
+++ b/crates/m68k/src/core/decode.rs
@@ -1193,29 +1193,25 @@ fn dispatch_group_4<B: AddressBus>(cpu: &mut CpuCore, bus: &mut B, opcode: u16) 
                                     return 20;
                                 }
                                 7 if cpu.is_040() => {
-                                    // 68040 access-error frame (30 words).
-                                    // The faulting instruction was rolled
-                                    // back at frame-build time, so a plain
-                                    // restart is the whole continuation --
-                                    // except the writeback protocol: a
-                                    // faulted write was pushed in slot 2,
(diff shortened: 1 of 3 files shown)


codewiz:
@codex review

LinuxJedi (approved):

PR #157, opened 2026-07-09
a4091: clear FIFOs on CTEST8.CLF strobe

codewiz:
A write to the 53C710 CTEST8 register with the CLF bit (0x04) drains
both FIFOs: the DMA FIFO (all four lanes) and the SCSI FIFO. The bit is
a self-clearing strobe, so it never latches in the register file (the
read path already masks it off; this makes the stored value agree).

This matches the datasheet semantics and the equivalent behavior in the
53C710 model I ported into Amiberry's `qemuvga/lsi53c710.cpp`. Note the
Amiberry side only zeroes its SCSI FIFO count because its DMA FIFO uses
a separate byte-count model; here both FIFOs are explicit, so both are
cleared.

Adds `ctest8_clf_drains_both_fifos`: fills every DMA lane and the SCSI
FIFO, strobes CLF, then verifies CTEST1 reads all-lanes-empty, the
SSTAT2 count is zero, and the CLF bit reads back clear. All 17 a4091
tests pass.

### DIFF
--- a/src/a4091.rs
+++ b/src/a4091.rs
@@ -42,6 +42,8 @@ const DIP_OFFSET: u32 = 0x008C_0003;
 /// 53C710 register-file byte offsets as the 68k sees them (the chip is
 /// wired big-endian on the A4091, so these are the driver's REG_ addresses).
 const REG_CTEST8: usize = 0x21;
+/// CTEST8.CLF: clear the DMA and SCSI FIFOs; self-clearing strobe.
+const CTEST8_CLF: u8 = 0x04;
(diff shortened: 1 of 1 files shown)


codewiz:
@codex review

codewiz:
Why were the checks canceled?

LinuxJedi:
> Why were the checks canceled?

Good question, was not me. Let's see what we can do about it.

LinuxJedi:
@codewiz must have been a GitHub bug, but there is a real failure now.

codewiz:
> @codewiz must have been a GitHub bug, but there is a real failure now.

Done. PTAL.

LinuxJedi (approved):

PR #173, opened 2026-07-13
launcher: add Host Mounts tab to the Machine Configuration GUI

codewiz:
image

Follow-up to the HOSTFS filesys work (#159): a config-screen tab to add and
edit host-directory mounts. Independent of the routing/FilesysUnit refactor
in #172 -- this touches only `src/video/*` and builds on `MountSpec` as it
landed in #159. The two can go in either order.

## What changed

- **Host Mounts tab**: per mount a directory (folder picker), a volume name,
  and a boot priority.
- Four editable slots (`FILESYS_GUI_SLOTS`). Mounts beyond that are preserved
  untouched on save rather than dropped, so a hand-written config with more
  entries survives a round trip through the GUI.
- Text entry now takes the platform layout-aware characters, so lowercase and
  underscores work in the volume-name and path boxes.
- Paths show in full, tail-clipped with a leading "..." when too long for the
  field.

## Testing

- `cargo test` green (1378 lib + 32 + 18 + 6), `cargo clippy --all-targets`
  clean, `cargo fmt --check` clean
- Manually exercised the tab in the launcher and saved a working config

🤖 Generated with [Claude Code](https://claude.com/claude-code) - and then hand-edited a bit

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1002,23 +1002,8 @@ impl Config {
                 if !m.path.is_dir() {
                     anyhow::bail!("[[filesys]] path {} is not a directory", m.path.display());
                 }
-                // The volume name becomes an AmigaDOS volume label (a DosList
-                // BSTR): 1-30 bytes, and no ':' '/' or NUL.
-                let vol = &m.volume;
-                if vol.is_empty() {
-                    anyhow::bail!("[[filesys]] volume name must not be empty");
-                }
-                if vol.len() > 30 {
-                    anyhow::bail!(
-                        "[[filesys]] volume name {vol:?} is too long ({} bytes; max 30)",
-                        vol.len()
-                    );
-                }
-                if vol.contains([':', '/', '\0']) {
-                    anyhow::bail!(
-                        "[[filesys]] volume name {vol:?} contains an invalid \
-                         character (no ':' '/' or NUL)"
-                    );
+                if let Some(err) = crate::filesys::volume_name_error(&m.volume) {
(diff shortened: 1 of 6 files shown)


codewiz:
Rebased on main (post-#172) and addressed the review:

- **Volume-name validation** (window.rs:3122): the checks were duplicated between the `[[filesys]]` validator and the `[scsi]`/`[ide]` drive-name parser, so I folded both into `filesys::volume_name_error()` and reused it in the launcher. Committing a name with ':' '/' NUL or over 30 bytes now shows the reason in the status line and keeps the field focused so it can be fixed, rather than saving a config that will not load. Net -27 lines in config.rs.
- **Round-trip test** (launcher.rs:966): added `host_mounts_round_trip_and_keep_entries_past_the_gui_slots`, covering >4 mounts, an untouched save, and a cleared slot; plus a test for the invalid-name path.
- **Slot renumbering** (launcher.rs:955): working as intended. `HOSTFS<n>` is defined as the position in the config, so clearing a slot renumbers the mounts after it exactly as deleting that `[[filesys]]` block from the TOML by hand would. The GUI slot labels are positional too, so the panel stays self-consistent after a reload. Making a cleared slot hold its number would need the mount to carry an explicit device name in the config, which is a separate change (and a reasonable one, if we want startup scripts to be able to rely on a stable HOSTFS number).

Tests, clippy and fmt are clean.

LinuxJedi (approved):
Looks great, many thanks!

PR #180, opened 2026-07-13
A3000/A4000: Ramsey, Super DMAC + WD33C93, Fat Gary, and register access logging

codewiz:
First slice of A3000/A4000 support. **Both machines now boot.**

## Ramsey

The memory controller of both big-box machines: a control register at `$DE0003` and a version register at `$DE0043` reporting `$0d` (Ramsey-04, A3000) or `$0f` (Ramsey-07, A4000).

The control register has to read back what was written. Guests write a mode and then spin until they see it, with interrupts and the MMU disabled -- a swallowed write is an unrecoverable hang, not a wrong value. cdhooper's `ziptest` froze the machine this way before this change.

Refresh timing is not modelled, so the MHz figure the diagnostic tools compute from it is meaningless (as it is under amiberry). The reset value describes the DRAM we actually claim, which is one place we are already more accurate: amiberry powers its control register up at zero and nothing programs it, so its A4000 reports a `256Kx4` configuration that was never fitted.

## `[debug] log_unmapped`

Logs CPU reads and writes that no device decodes, within a hex range or `all`. Reads report the floating bus value they returned; writes report the value that went nowhere.

A missing register is normally invisible -- a read floats, a write is dropped, and the guest sulks or hangs with no diagnostic. This makes it a two-minute job instead of a bisect. It found every bug in this PR. It replaces the `COPPERLINE_DIAG_GAYLE` float logging, which was an env flag hardcoded to `$D80000-$F00000` and named after the wrong chip.

## Fat Gary

The bus controller in front of Ramsey, and the reason the diagnostic tools could not see the Ramsey we already emulated correctly.

Gary and Ramsey share one address decode, and it is cruder than a register map suggests: the byte lane picks the chip (lanes 0-2 are Gary, lane 3 is Ramsey), two address bits pick Ramsey's control vs version register, and the whole thing mirrors every `$100` across the page. So the version register Kickstart reads at `$DE0043` answers equally at `$DE0047` or `$DE0143`, and the addresses the tools appear to read at random are mirrors.

Gary's three registers are single bits in bit 7: TIMEOUT (`$DE0000`, BERR vs DSACK on an unanswered cycle), TOENB (`$DE0001`), and COLDBOOT (`$DE0002`). The last one is read/write, which is what makes it a Gary detector -- xSysInfo writes bit 7, reads a custom register to defeat the sticky bus, and expects to read bit 7 back. Modelling Ramsey as two fixed addresses left all of that floating, so xSysInfo concluded there was no Fat Gary and then, by its own logic, never looked for the Ramsey behind one:

```c
void detect_ramsey(void) {
    if (hw_info.gary_type != FAT_GARY) { // no fat gary -> no ramsey!
        hw_info.ramsey_rev = 0;
        return;
    }
```

Neither register changes emulated behaviour -- bus timeouts are not modelled and an unanswered cycle floats rather than faulting -- but they are what lets the machine identify as what it is. It also gives Kickstart's `$DE0000` writes, dropped on every boot until now, somewhere to land.

Gary is a `GateArray` variant rather than a separate flag. A machine has exactly one bus gate array, Gayle and Gary both decode the `$DE0000` page, and fitting both would make the decode ambiguous; the A3000/A4000 profiles now say `FatGary` instead of claiming no gate array at all.

## A3000 and A4000 profiles

Ramsey, Gary instead of Gayle, 2 MB chip RAM, 68030/68040 at 25 MHz. No motherboard fast RAM: the OS is happy taking fast RAM from the Zorro III board, so `[memory] z3` covers it.

## Super DMAC + WD33C93

The A3000's motherboard SCSI at `$DD0000`. The SDMAC is the successor of the A2091's DMAC and pairs with the same WD33C93, so `crate::scsi`'s chip core drives it unchanged and `src/sdmac.rs` is a second front-end onto it beside `src/a2091.rs` -- the layering amiberry uses too, where both controllers share one `wd_state`.

What differs from the DMAC: byte-wide registers instead of word-wide, a 32-bit DMA address counter instead of Zorro II's 24-bit one, the SSPB serial-bus registers, and no boot ROM at all -- Kickstart's own `scsi.device` drives it, so the machine either has a working controller or it hangs in device init. The SDMAC-04 also dropped the word transfer counter (the count lives in the WD33C93), and reading it back as zero is exactly how both Kickstart and cdhooper's `sdmac` identify the part.

`SASR`/`SCMD` are decoded at their `$41`/`$47` aliases as well as `$49`/`$43`. That is not cosmetic: SysInfo polls the auxiliary status through the `$41` alias, and before it decoded it read the floating bus 250,075 times and hung the machine with the power LED flashing.

## State of the two machines

- **A3000 boots**, on Kickstart's own `scsi.device` against a real SDMAC + WD33C93. No drives are attached to it yet, so boot from a Zorro controller or a `[[filesys]]` mount.
- **A4000 boots**, but slowly: its IDE at `$DD2020` is not implemented, so `scsi.device` probes it and waits out a timeout on every boot. The log shows the probe exactly -- a write of `$A0` to the device/head register at `$DD203A`, then 138 status reads at `$DD203E` that never come back ready.

## Next

1. A4000 IDE at `$DD2020`, reusing the ATA core in `gayle.rs`. Its byte-swap has to become a property rather than a constant: Gayle swaps the IDE data bus, the A4000 does not.
2. Drives on the A3000's SCSI bus (`[scsi]` currently only fits the Zorro controllers).

## Testing

`cargo test` (1406 lib tests), clippy clean.

Booted an A4000 (68040) and ran cdhooper's `ziptest`: reports `Ramsey-07 $f $08`, `1Mx4 (4MB per bank)` -- correct, where amiberry reports RAM that was never fitted.

Booted an A3000 (68030) to Workbench and ran cdhooper's `sdmac`, which passes its Ramsey and SDMAC tests and identifies the part as `SDMAC-04`. amiberry fails four of those same subtests (ACR at `$DD000C` reads 0, SSPBDAT at `$DD0058` reads `$ff`, an invalid WD register read should float `$ff`, and its SDMAC-04 identification is a false positive off an unimplemented WTC), so this was implemented against cdhooper's tests rather than against amiberry's behaviour.

Both SysInfo and xSysInfo identify the machine: Fat Gary rev `$0`, Ramsey-04 rev `$D`, SDMAC, ECS Agnus/Denise, 68030 with the MMU in use. Neither would report the Ramsey before Gary existed -- xSysInfo's `detect_ramsey()` returns early unless it first finds a Fat Gary, and it finds one by writing and reading back bit 7 of the coldboot register.

With `[debug] log_unmapped` pointed at `$DD0000-$DEFFFF`, a full A3000 boot now produces no floating accesses at all.

### DIFF
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -104,8 +104,11 @@ model = "68000"
 # A500 (ECS 8372A Agnus, OCS 8362 Denise, 512 KiB chip + 512 KiB trapdoor
 # slow RAM) -- the most common and most-targeted Amiga.
 # The big-box profiles (A3000, A4000) are new and incomplete: Ramsey and the
-# chipset are there, but their on-board mass storage is not (the A3000 Super
-# DMAC SCSI and the A4000 IDE), so they need a Zorro controller to boot from.
+# chipset are there, but their on-board mass storage is not. The A4000 boots
+# (Kickstart times out probing the missing IDE at $DD2020, so give it a Zorro
+# controller or a [[filesys]] mount to boot from); the A3000 does not boot yet,
+# because its Super DMAC has no WD33C93 behind it and scsi.device waits forever
+# for an interrupt from it.
 # [machine]
 # profile = "A600"      # A1000 | A500 | A500OCS | A500Plus | A600 | A1200 | A3000 | A4000 | CDTV | CD32
 # rtc = true            # add a $DC0000 battery RTC (default: only A500+/CDTV/A3000/A4000 have one)
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -103,12 +103,12 @@ model = "68000"
 # Without any [machine] section the default machine is that same Rev 6A
 # A500 (ECS 8372A Agnus, OCS 8362 Denise, 512 KiB chip + 512 KiB trapdoor
 # slow RAM) -- the most common and most-targeted Amiga.
-# The big-box profiles (A3000, A4000) are new and incomplete: Ramsey and the
-# chipset are there, but their on-board mass storage is not. The A4000 boots
-# (Kickstart times out probing the missing IDE at $DD2020, so give it a Zorro
-# controller or a [[filesys]] mount to boot from); the A3000 does not boot yet,
-# because its Super DMAC has no WD33C93 behind it and scsi.device waits forever
-# for an interrupt from it.
+# The big-box profiles (A3000, A4000) are new and incomplete, but both boot.
+# The A3000 has its motherboard SCSI (Super DMAC + WD33C93 at $DD0000, driven by
+# Kickstart's own scsi.device) with no drives attached yet; the A4000's IDE at
+# $DD2020 is not emulated at all, so Kickstart spends several seconds probing for
+# it on every boot. Give either machine a Zorro controller or a [[filesys]] mount
+# to boot from.
 # [machine]
 # profile = "A600"      # A1000 | A500 | A500OCS | A500Plus | A600 | A1200 | A3000 | A4000 | CDTV | CD32
 # rtc = true            # add a $DC0000 battery RTC (default: only A500+/CDTV/A3000/A4000 have one)
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -122,7 +122,7 @@ model = "68000"
 # space to make "all" a firehose (a million lines a boot), so prefer aiming it
 # at one window.
 # [debug]
-# log_unmapped = "DD0000-DE0000"   # hex START-END, end exclusive; or "all"
+# log_unmapped = "DD0000-DEFFFF"   # hex START-END, end included; or "all"
 
 
 # Extended ROM image for CDTV/CD32 machines: a 256 KiB image maps at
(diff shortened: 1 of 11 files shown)


codewiz:
I got the A3000 profile to boot (with a hack to disable the scsi.device resident tag in the Kickstart) and pass some of the amiga_sdmac tests:

image

xSysInfo still does not identify Ramsey. SysInfo says it's revision $D and then crashes (infinite power LED flashes):

image

Also, something seems to be off with ECS emulation - separate bug.

codewiz:
Are you ok with the "log_unmapped" feature as designed or should I pull it out before merging this series?

It's been invaluable in debugging Ramsey & SDMAC, but you might prefer a design where each chip logs invalid register access separately. I'm happy to pull it out or rework, but I'd rather not block A3000/A4000 support on a debug feature.

See commit 5e6ab7dd486685761b161e1f46afd00b53f7742e for the details.

LinuxJedi (approved):
I think this lays great groundwork, and the new logging method is a good idea. We can always expand on it if it isn't enough later.

PR #184, opened 2026-07-14
scsi: put drives on the A3000 motherboard SCSI bus

codewiz:
The A3000's Super DMAC and its WD33C93 landed in #180, but nothing could be
plugged into them: `[scsi]` only fitted the Zorro controllers, so a big-box
machine still needed an A2091/A4091 or a `[[filesys]]` mount to boot.

`[scsi] controller = "a3000"` now fits the drives to the SDMAC's WD33C93. It is
the default controller on a machine that has a Super DMAC, so on an A3000 a bare

    [machine]
    profile = "A3000"
    [scsi]
    unit0 = "workbench.hdf"

is enough. Unlike the Zorro boards it takes no boot ROM - Kickstart's own
scsi.device drives it - so the config rejects `rom =` there, and rejects the
controller entirely on a machine without the silicon. A Zorro board still fits
an A3000 if you name it explicitly.

The launcher's SCSI picker gains an "A3000 (onboard)" entry, offered only where
the hardware is, with the ROM row greyed out. Switching the profile away from
the A3000 moves a fitted onboard controller back to the A2091, so the drives
survive the change rather than failing validation on Run.

The SDMAC and WD33C93 emulation needed no changes.

Tested by booting an A3000 with an RDB HDF on unit 0 and nothing else to boot
from: Kickstart selects the target, reads the RDB, mounts and autoboots to
Workbench.

### DIFF
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -104,11 +104,11 @@ model = "68000"
 # A500 (ECS 8372A Agnus, OCS 8362 Denise, 512 KiB chip + 512 KiB trapdoor
 # slow RAM) -- the most common and most-targeted Amiga.
 # The big-box profiles (A3000, A4000) are new and incomplete, but both boot.
-# The A3000 has its motherboard SCSI (Super DMAC + WD33C93 at $DD0000, driven by
-# Kickstart's own scsi.device) with no drives attached yet; the A4000's IDE at
-# $DD2020 is not emulated at all, so Kickstart spends several seconds probing for
-# it on every boot. Give either machine a Zorro controller or a [[filesys]] mount
-# to boot from.
+# The A3000 boots off its motherboard SCSI (Super DMAC + WD33C93 at $DD0000,
+# driven by Kickstart's own scsi.device): put drives on it with [scsi]. The
+# A4000's IDE at $DD2020 is not emulated at all, so Kickstart spends several
+# seconds probing for it on every boot; give that machine a Zorro controller or
+# a [[filesys]] mount to boot from.
 # [machine]
 # profile = "A600"      # A1000 | A500 | A500OCS | A500Plus | A600 | A1200 | A3000 | A4000 | CDTV | CD32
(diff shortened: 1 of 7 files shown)


codewiz:
Good catch on rom_odd - fixed in 6206c40. It is the odd half of a split even/odd EPROM dump, so it is meaningless without rom, and the config validator rejects that pair; clearing the even half in the launcher used to leave the odd half behind and fail on Run. The launcher now only writes rom_odd when rom is set.

codewiz:
Now we pass 100% of cdhopper's sdmac test:

```
WB:System/Diag/sdmac
Memory controller:   Ramsey-04 $d
Ramsey config:       256Kx4, 154 clock refresh
SCSI DMA Controller: SDMAC-04
SCSI Controller:     WD33C93A 00-04 or 00-03 microcode 00, 8.7 MHz
WDC Configuration:   Polled Mode, 0 msec timeout, Async

Ramsey test:  PASS
SDMAC test:   PASS
WDC test:     PASS
3.WB:>
```

codewiz:
Both new review comments are already handled:

- rom_odd without rom: fixed in 6206c40 (the comment was posted against the original diff). The launcher no longer writes rom_odd unless rom is set.
- `SCSI_CONTROLLERS.into_iter()` and `*c`: this one is a false positive. `Iterator::filter` passes `&Item` to its closure whatever the iterator yields, so the closure parameter is `&Option<ScsiController>` and the deref is needed. It compiles, and the launcher tests pass.

Also pushed d21fec0: the WD33C93 register file ends at the data register ($19). Registers $1A-$1E are not registers and float $ff, and $1F is the auxiliary status aliased in read-only. Both were plain RAM here, so cdhooper sdmac tool failed its WDC test and reported "SCSI Controller: Not detected: INVALID" - it identifies the chip by checking exactly those two properties. It now passes all three tests on an emulated A3000 and reports "WD33C93A 00-04 or 00-03 microcode 00, 8.7 MHz".

LinuxJedi (approved):

PR #210, opened 2026-07-17
Support any controller device in either game port

LinuxJedi:
## Summary

A real Amiga accepts a mouse, digital joystick, CD32 pad, or analogue controller in either game port; Copperline hardwired mouse=port 1 and joystick/CD32 pad=port 2 across every layer. This models the ports as pluggable devices end to end.

### Core
- `InputState` becomes `ports: [ControllerPort; 2]` carrying a `PortDevice` (`mouse`/`joystick`/`cd32`/`analogue`/`none`) plus every line a controller can drive. The JOYxDAT counters are chip-side, so they survive device swaps and JOYTEST loads them whatever is plugged in. `STATE_VERSION` 31.
- The CD32 pad serial protocol works on both ports: POT0X mode select (POTGO bits 9/8), /FIR0 clock, POT0Y data mirror the port-2 wiring, with an independent shifter per port.
- An analogue device presents pot resistances as the exact inverse of the RC comparator threshold (position N latches POTxDAT count N), paddle buttons on the left/right lines per the HRM, pots seeded centred on plug-in. POT0DAT/POT1DAT join the debugger's side-effect-free latch view.
- Machine resets (keyboard chord, CPU RESET, cold boot) keep the plugged devices and analogue knob positions; the driven lines release and the chip counters clear.

### Surface
- `[input] port1/port2` + `--port1`/`--port2` (stock wiring by default; the CD32 profile supplies its bundled pad, subsuming the `cd32_pad` bool).
- Scripted input takes an optional trailing PORT token (`joy-after SECS BTN MS [PORT]`) plus a new `pot-after SECS X Y [PORT]`; omitted, every flag keeps its traditional port, so existing scripts, recordings, TOMLs, and CCP clients are unchanged.
- The recorder diffs each port by its device, emitting port tokens only when non-default: default-wiring sessions record byte-identically, other wirings note a `# ports:` header.
- CCP: `input.mouse`/`input.joy` gain an optional `port`, plus `input.analogue`, `input.set_port` (hot-plug), `input.get_ports`.
- Host routing: the mouse drives the lowest-numbered mouse port; with joysticks in both ports the gamepad and keyboard mapping drive one each (two-player), the input mode picking which gets the lower port. The launcher gains an Input tab; the runtime menu hot-plugs devices per port, clearing stale scripted joy ownership.
- Gamepad-analogue-stick-to-pot mapping is deferred (needs a gamepads.toml calibration schema extension); analogue is script/CCP/config-driven for now, noted in the docs.

## Verification

All headless against AmigaTestKit 1.21 (F4 controller-port test), plus `cargo test` (1640 tests), `clippy`, `fmt`, and the golden-probe suite in release:

- **Defaults regression**: the F4 screenshot is byte-identical to the pre-change render.
- **Swapped wiring** (`--port1 joystick --port2 mouse`): port-tagged input lands on each port's view.
- **CD32 pad in port 1**: ATK reports *Pad Detected* with exactly the held buttons (red, green, rewind) lit; a mouse in a port correctly reads *Pad Not Detected*.
- **Analogue**: ATK's Analog view actively strobes POTGO and reads back `--pot-after 50 200` as literally (50,200), min/max tracing the sweep from the centre seed.
- **Reset survival**: input keeps driving a swapped port across a scripted Ctrl-Amiga-Amiga reset.
- **Recorder round-trip**: a recorded swapped-wiring session replays to a byte-identical screenshot.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,9 +70,10 @@ repeat.
 |---|---|
 | `--press-after SECS KEY` | Press and release an Amiga key (~100 ms hold) |
 | `--key-after SECS KEY MS` | Hold a key for exactly MS milliseconds |
-| `--click-after SECS BUTTON MS` | Mouse button (`left`/`right`/`middle`) for MS ms |
-| `--joy-after SECS BUTTON MS` | Port-2 joystick / CD32-pad control (`up`/`down`/`left`/`right`/`red`/`blue`/...) |
-| `--mouse-after SECS DX DY` | Relative mouse motion |
+| `--click-after SECS BUTTON MS [PORT]` | Mouse button (`left`/`right`/`middle`) for MS ms (default port 1) |
+| `--joy-after SECS BUTTON MS [PORT]` | Joystick / CD32-pad control (`up`/`down`/`left`/`right`/`red`/`blue`/...) (default port 2) |
+| `--mouse-after SECS DX DY [PORT]` | Relative mouse motion (default port 1) |
+| `--pot-after SECS X Y [PORT]` | Analogue stick/paddle position, 0-255 per axis (default port 2) |
 | `--insert-disk-after SECS DFN PATH` | Insert a disk image into `df0`..`df3` |
 | `--script FILE` | Same directives from a file, one per line, no leading dashes |
 | `--record-input PATH` | Record all machine-bound input as a replayable script |
@@ -80,6 +81,13 @@ repeat.
 `KEY` is a raw key code (`0x45`) or a name (`ctrl`, `f1`, `esc`, letters,
 digits). A session played by hand under `--record-input` (or Cmd+Shift+R /
 Alt+Shift+R in the window) replays deterministically via `--script`.
+
+Either controller port takes any device -- `[input] port1/port2` in the
+TOML, or `--port1`/`--port2` (`mouse`/`joystick`/`cd32`/`analogue`/`none`;
+default mouse + joystick, CD32 pad on the CD32 profile). The scripted-input
+flags' optional trailing `PORT` token (`1` or `2`) aims an event at either
+port; omitted, each flag keeps its traditional port, so existing scripts
+are unchanged.
 
 ## Save states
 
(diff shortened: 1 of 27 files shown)


LinuxJedi:
Follow-up pushed for the two routing gaps:

**1. Two joysticks, no gamepad**: a second keyboard mapping on the numeric keypad (`8`/`2`/`4`/`6` directions, `0` fire, `.` second button, numpad Enter play) now stands in for the gamepad whenever no physical pad is present in a two-controller setup, so both ports are drivable from one keyboard. The layout is collision-free against the cursor mapping's CD32 letters (C/X/D/S/Z/A), which is why it lives on the numpad rather than WASD.

**2. Two mice / keyboard mouse**: the keyboard mapping is now device-aware and drives whatever its routed port carries. On a mouse port the same keys become the pointer: cursor keys move it (~150 counts/s), fire keys = left button, `X` = right, `D` = middle. With two mice, the host mouse takes the lower-numbered port and the cursor-key mapping drives the second in `keyboard` mode; in `gamepad` mode the second mouse is undriven and the keyboard keeps passing through, preserving the no-surprise default.

Routing is one `HostRouting` struct per quantum with unchanged precedence (pad > scripted joy state > keyboard mapping). Stock wiring is untouched — the AmigaTestKit defaults render is still byte-identical — and the new paths are unit-tested through the real input pump.

LinuxJedi:
The Input tab now explains itself: a live summary under the rows states which host input drives each port, recomputed as the values cycle.

For the two-joystick example:

```
With these settings:
  Port 1 is driven by the gamepad (numpad keys without a pad)
  Port 2 is driven by cursor keys (Right Ctrl / Right Alt = fire)
```

Other wirings get the matching lines ("the host mouse", "cursor keys as a mouse (fire keys = buttons)", "--pot-after scripting or the control protocol"), and undriven combinations name the remedy ("nothing (flip Joystick input to keyboard)"). The lines are generated by the same `host_routing_for` function the runtime input pump uses — extracted so the GUI's promise can't drift from the actual routing — with a test asserting the wording per configuration and that every device/mode combination fits the pane, plus a `launcher-input` render in the UI preview suite.

Issue #19, opened 2026-06-21
RTG support


LinuxJedi:
General reminder for me to add RTG support

codewiz:
I'm giving a stab at this.

Rather than implementing the crufty and fictictious uaegfx, I decided to emulate open hardware that you can obtain today. This essentially means PiStorm, ZZ9000 or Z3660.

The ZZ9000 driver is high quality and uses a regular Zorro card with MMIO registers. Various operations are accelerated.
The Z3660 driver is derived from the ZZ9000 and seems slightly easier to emulate, so I'll try this first.

codewiz:
Making good progress:

<img width="1098" height="962" alt="Image" src="https://github.com/user-attachments/assets/ffe0cc3b-32b5-4df0-ae74-0dbf65c97bb4" />
<img width="1098" height="962" alt="Image" src="https://github.com/user-attachments/assets/3df8f0ab-0c6a-432c-83c4-555402e9fb44" />
<img width="1099" height="963" alt="Image" src="https://github.com/user-attachments/assets/84ee9a82-f0ff-47a2-b652-9e7c2eb24bd5" />

codewiz:
My Workbench renders in 16bpp and 24bpp (with glitches):

<img width="1101" height="965" alt="Image" src="https://github.com/user-attachments/assets/1629cf4e-5272-453d-87bc-6baf7c001d04" />

codewiz:
Reached MVP level. I'll push a PR shortly.

<img width="1528" height="1301" alt="Image" src="https://github.com/user-attachments/assets/164e0b3c-5c11-4bb1-ac6e-bd816f5a0129" />

codewiz:
Tested with this scratch build of the Z3660 driver from git commit af0733d2279fae3f0b79a731b837ba04f3e31718:

[Z3660-P96-20260717.zip](https://github.com/user-attachments/files/30133441/Z3660-P96-20260717.zip)

You can also use the official distribution, but in order to switch between AGA and Z3660 modes you should set the tooltype DISPLAYCHAIN=YES (and delete the duplicate entry in the icon).

codewiz:
PR under review here: https://github.com/LinuxJedi/Copperline/pull/212

codewiz:
I asked Claude to analyze the entire video pipeline from Z3660 framebuffer to screen and find where  destructive resampling happens with high resolution modes:

---

[...]

PR #212, opened 2026-07-17
RTG: Emulate the Z3660 video board

codewiz:
Image

## Summary

Emulates the Z3660 accelerator's FPGA RTG core as a Zorro III autoconfig
board (manufacturer 0x144B, product 1, one 128 MB window), driven by the
unmodified open-source Z3660.card Picasso96 driver. Workbench runs on
Z3660 screen modes in 8/16/32-bit with a working mouse pointer, and the
display switches cleanly between the native chipset and the RTG
framebuffer.

First RTG board for #19. The ZZ9000 (largely register-compatible with
this board) and a Copperline-native card discoverable by the bundled
AROS ROM's p96gfx HIDD are natural follow-ups on the same
infrastructure.

Enabled with:

```toml
[z3660]
enabled = true
```

## What is emulated

- Autoconfig identity and the 32-bit register file at window offsets
  0x100-0x7FF. Every register access and mailbox op is logged by name,
  so the board doubles as a driver-development rig.
- FindCard/InitCard probes: FW_VERSION (reports v1.06), INT_STATUS,
  MONITOR_SWITCH (0 = no CIA display-switch fiddling).
- VBLANK_STATUS follows a 60 Hz frame phase in emulated colour clocks,
  so WaitVerticalSync/GetVSyncState behave like hardware.
- SetGC/SetPanning: custom modelines (CUSTOM_VIDMODE parameter pairs),
  MODE (vmode/colormode/scalemode), ORIG_RES, and the PAN op.
- Scanout: the panned VRAM framebuffer is composed in all four pixel
  formats (CLUT through the captured palette, R5G6B5, R5G5B5,
  B8G8R8A8) and presented by the window and the CCP capture path;
  OP_CAPTUREMODE switches back to native chipset output.
- The GFXData mailbox blitter ops, executed synchronously on the
  doorbell write as the bus-stalling hardware does: FILLRECT, COPYRECT
  (overlap-safe, 8-bit plane mask), COPYRECT_NOMASK, INVERTRECT,
  RECT_TEMPLATE and RECT_PATTERN (JAM1/JAM2/COMPLEMENT + INVERSVID),
  DRAWLINE (patterned, with the COMPLEMENT XOR mode the shell cursor
  uses), and the planar P2C/P2D blits that render icons and window
  gadget imagery.
- The hardware sprite: the mouse pointer image, colours, and position
  ride the sprite ops and are composited over the output at present
  time, never touching VRAM -- like the real output-stage overlay.

Semantics follow the board's ARM firmware (z3660-firmware, GPL-3.0
like Copperline), including its quirks: the GFXData pitch field is
longwords for fill/copy/invert/line but raw bytes for
template/pattern, colours ride in guest byte order (the little-endian
ARM reads the 68k values byteswapped), and minterms are the firmware's
0-15 enum, not raw blitter bytes.

Not implemented yet (logged when hit): the ACC_OP surface ops, exotic
minterms, screen-split, overlays, and the video-capture path. RTG
frames are currently scaled into the Amiga-sized presentation texture;
pixel-perfect presentation is a follow-up.

## Test plan

- Unit tests for the register latch, palette capture, scanout in all
  pixel formats, pan, each blit-op family, the planar decodes, and the
  sprite compositing cycle (show/upload/colour/move/hide).
- Live: Workbench 3.2 on Z3660:800x600 and 1920x1200 modes in 8, 16,
  and 32-bit with the real Z3660.card 1.06 built from source; window
  dragging, shell text and cursor, icons, and gadget imagery render;
  ScreenMode test-cycle and DISPLAYCHAIN=YES switch RTG <-> AGA both
  ways. Headless: deterministic boots driving the driver via the
  serial shell with CCP screenshots.
- cargo test (1593), clippy, fmt clean.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/video/present_common.rs
+++ b/src/video/present_common.rs
@@ -18,17 +18,18 @@ pub const fn rgba(r: u32, g: u32, b: u32) -> u32 {
     0xFF00_0000 | (b << 16) | (g << 8) | r
 }
 
-/// Compose the active RTG board frame (Z3660 scanout) into an
-/// FB_WIDTH-stride presentation buffer: the board frame lands in `scratch`
-/// at its own resolution and is scaled horizontally (nearest) into `out`,
-/// one output row per board row -- the shared vertical scaling then maps
-/// rows to the output height like any chipset frame. Returns the row count,
-/// or `None` while no RTG board drives the display.
+/// Compose the active RTG board frame (Z3660 scanout). The board frame lands
+/// in `scratch` at its native resolution; `out` gets an FB_WIDTH-stride copy
+/// (nearest, one output row per board row) for the screenshot path, which
+/// reads the shared presentation buffer. Returns
+/// `(present_rows, native_w, native_h)`, or `None` while no RTG board drives
+/// the display. The window presents the native `scratch` frame directly (see
+/// [`crate::video::window`]); `out` exists only for the screenshot path.
 pub fn compose_rtg_present(
     bus: &crate::bus::Bus,
     scratch: &mut Vec<u32>,
     out: &mut Vec<u32>,
-) -> Option<usize> {
+) -> Option<(usize, u32, u32)> {
     let (w, h) = bus.rtg_frame(scratch)?;
     let (w, h) = (w as usize, h as usize);
     out.clear();
@@ -40,7 +41,7 @@ pub fn compose_rtg_present(
             *px = src[x * w / FB_WIDTH];
         }
     }
-    Some(h)
+    Some((h, w as u32, h as u32))
 }
 
 pub const STANDARD_PAL_VISIBLE_WIDTH: usize = 320 * 2;
--- a/src/video/present_common.rs
+++ b/src/video/present_common.rs
@@ -38,7 +38,11 @@ pub fn compose_rtg_present(
         let src = &scratch[y * w..(y + 1) * w];
         let dst = &mut out[y * FB_WIDTH..(y + 1) * FB_WIDTH];
         for (x, px) in dst.iter_mut().enumerate() {
-            *px = src[x * w / FB_WIDTH];
+            // Sample the centre of each output pixel's source span, not its
+            // left edge: `x * w / FB_WIDTH` tops out below `w - 1` whenever
+            // the board frame is wider than FB_WIDTH, so the rightmost
+            // source columns never reach a downscaled screenshot.
+            *px = src[(2 * x + 1) * w / (2 * FB_WIDTH)];
         }
     }
     Some((h, w as u32, h as u32))
--- a/src/video/present_common.rs
+++ b/src/video/present_common.rs
@@ -18,6 +18,31 @@ pub const fn rgba(r: u32, g: u32, b: u32) -> u32 {
     0xFF00_0000 | (b << 16) | (g << 8) | r
 }
 
+/// Compose the active RTG board frame (Z3660 scanout) into an
+/// FB_WIDTH-stride presentation buffer: the board frame lands in `scratch`
+/// at its own resolution and is scaled horizontally (nearest) into `out`,
+/// one output row per board row -- the shared vertical scaling then maps
+/// rows to the output height like any chipset frame. Returns the row count,
+/// or `None` while no RTG board drives the display.
+pub fn compose_rtg_present(
+    bus: &crate::bus::Bus,
+    scratch: &mut Vec<u32>,
+    out: &mut Vec<u32>,
+) -> Option<usize> {
+    let (w, h) = bus.rtg_frame(scratch)?;
+    let (w, h) = (w as usize, h as usize);
+    out.clear();
+    out.resize(h * FB_WIDTH, 0xFF00_0000);
+    for y in 0..h {
+        let src = &scratch[y * w..(y + 1) * w];
+        let dst = &mut out[y * FB_WIDTH..(y + 1) * FB_WIDTH];
+        for (x, px) in dst.iter_mut().enumerate() {
+            *px = src[x * w / FB_WIDTH];
+        }
+    }
+    Some(h)
+}
+
 pub const STANDARD_PAL_VISIBLE_WIDTH: usize = 320 * 2;
 pub const STANDARD_PAL_VISIBLE_LINES: usize = 256;
 pub const STANDARD_PAL_VISIBLE_START_VPOS: u32 = 0x2C;
--- a/src/z3660.rs
+++ b/src/z3660.rs
@@ -15,9 +15,9 @@
 //! is honest memory. The display pipeline presents the panned framebuffer
 //! (all four pixel formats, palette captured from the upload stream) when
 //! the driver switches the display to RTG, and the core blitter ops
-//! (fill/copy/invert/template/pattern) execute into VRAM on the doorbell.
-//! Still stubbed: DRAWLINE, P2C/P2D, the ACC_OP surface ops, and the
-//! hardware sprite (the pointer is invisible until it is composited).
+//! (fill/copy/invert/template/pattern) execute into VRAM on the doorbell,
+//! with the hardware sprite (mouse pointer) composited over the output.
+//! Still stubbed: DRAWLINE, P2C/P2D, and the ACC_OP surface ops.
 
 use crate::zorro_device::{DeviceHost, ZorroDevice};
 
@@ -41,6 +41,7 @@ const BACKED_BYTES: usize = 0x0400_0000;
 
 // The registers the FindCard/init path probes.
 const REG_MODE: u32 = 0x100;
+const REG_SPRITE_BITMAP: u32 = 0x174;
 const REG_VBLANK_STATUS: u32 = 0x17C;
 const REG_BLITTER_DMA_OP: u32 = 0x180;
 const REG_ACC_OP: u32 = 0x184;
@@ -110,6 +111,17 @@ pub struct Z3660 {
     /// Whether SetGC has programmed a display mode since power-on; RTG
     /// scanout is meaningless before the first mode set.
     mode_set: bool,
+    /// Hardware sprite (the mouse pointer): composited over the frame at
+    /// scanout, never written to VRAM. Pixel values 0-3 index
+    /// `sprite_colors` (0 = transparent), row-major `sprite_w` wide.
+    sprite_visible: bool,
+    sprite_x: i32,
+    sprite_y: i32,
+    sprite_w: usize,
+    sprite_h: usize,
+    sprite_pix: Vec<u8>,
+    /// Sprite pens as 0x00RRGGBB (pen 0 unused).
+    sprite_colors: [u32; 4],
 }
 
 impl Z3660 {
@@ -123,6 +135,13 @@ impl Z3660 {
             pan_offset: 0,
             pan_width: 0,
             mode_set: false,
+            sprite_visible: false,
+            sprite_x: 0,
+            sprite_y: 0,
+            sprite_w: 0,
+            sprite_h: 0,
+            sprite_pix: Vec::new(),
+            sprite_colors: [0; 4],
         }
     }
 
@@ -202,6 +221,15 @@ impl Z3660 {
                     value & 0xFFFF
                 );
             }
+            REG_SPRITE_BITMAP => {
+                // SetSprite: 1 = show the hardware sprite (2 = hide; the
+                // driver ships only the show write, hiding via position).
+                if value == 1 {
+                    self.sprite_visible = true;
+                } else if value == 2 {
+                    self.sprite_visible = false;
+                }
+            }
             REG_CUSTOM_VIDMODE => self.vidmode_param = value,
             REG_CUSTOM_VIDMODE_DATA => {
                 log::info!(
@@ -271,6 +299,9 @@ impl Z3660 {
         const OP_RECT_TEMPLATE: u32 = 5;
         const OP_RECT_PATTERN: u32 = 6;
         const OP_INVERTRECT: u32 = 9;
+        const OP_SPRITE_XY: u32 = 11;
+        const OP_SPRITE_COLOR: u32 = 12;
+        const OP_SPRITE_BITMAP: u32 = 13;
         const MINTERM_SRC: u8 = 0xC0;
         const JAM1: u8 = 0;
         const INVERSVID: u8 = 8;
@@ -438,6 +469,45 @@ impl Z3660 {
                     }
                 }
             }
+            OP_SPRITE_XY => {
+                self.sprite_x = i32::from(self.be16(g + 0x10) as i16);
+                self.sprite_y = i32::from(self.be16(g + 0x18) as i16);
+            }
+            OP_SPRITE_COLOR => {
+                // rgb0 was assembled bytewise on the 68k as B,G,R,0; the
+                // u8offset byte selects the pen (driver sends idx+1).
+                let pen = (self.vram[g + 0x3B] & 3) as usize;
+                let (b, gr, r) = (rgb0 >> 24, (rgb0 >> 16) & 0xFF, (rgb0 >> 8) & 0xFF);
+                self.sprite_colors[pen] = (r << 16) | (gr << 8) | b;
+            }
+            OP_SPRITE_BITMAP => {
+                // Amiga 2-plane sprite rows at offset[1]: `hires` words of
+                // plane 0 then plane 1 per row; pixel value = plane bits.
+                let (w, h) = (x1, y1);
+                let hires = y2 + 1;
+                // Row: `hires` words of plane 0, then plane 1 (the driver
+                // copies (w>>3)*2*hires bytes per row).
+                let row_bytes = (w / 8) * 2 * hires;
+                if w == 0 || h == 0 || w > 16 * hires {
+                    log::debug!("z3660: sprite bitmap {w}x{h} hires {hires} unsupported");
+                    self.sprite_pix.clear();
+                    return;
+                }
+                self.sprite_w = w;
+                self.sprite_h = h;
+                self.sprite_pix = vec![0u8; w * h];
+                for y in 0..h {
+                    let row = src + y * row_bytes;
+                    for x in 0..w {
+                        let word = x / 16;
+                        let bit = 15 - (x & 15);
+                        let p0 = self.be16(row + 2 * word);
+                        let p1 = self.be16(row + 2 * hires + 2 * word);
+                        let v = (((p1 >> bit) & 1) << 1) | ((p0 >> bit) & 1);
+                        self.sprite_pix[y * w + x] = v as u8;
+                    }
+                }
+            }
             _ => {
                 log::debug!("z3660: dma-op {} not executed yet", dma_op_name(op));
             }
@@ -531,6 +601,26 @@ impl Z3660 {
                     _ => (self.vram[at + 2], self.vram[at + 1], self.vram[at]),
                 };
                 out.push(0xFF00_0000 | (u32::from(b) << 16) | (u32::from(g) << 8) | u32::from(r));
+            }
+        }
+        // The hardware sprite (mouse pointer) is composited over the output
+        // stream; it never exists in VRAM. Pen 0 is transparent.
+        if self.sprite_visible && !self.sprite_pix.is_empty() {
+            for sy in 0..self.sprite_h {
+                let py = self.sprite_y + sy as i32;
+                if py < 0 || py >= h as i32 {
+                    continue;
+                }
+                for sx in 0..self.sprite_w {
+                    let px = self.sprite_x + sx as i32;
+                    let pen = self.sprite_pix[sy * self.sprite_w + sx];
+                    if pen == 0 || px < 0 || px >= w as i32 {
+                        continue;
+                    }
+                    let rgb = self.sprite_colors[pen as usize];
+                    out[py as usize * w as usize + px as usize] =
+                        0xFF00_0000 | ((rgb & 0xFF) << 16) | (rgb & 0xFF00) | ((rgb >> 16) & 0xFF);
+                }
             }
         }
         Some((w, h))
@@ -628,6 +718,9 @@ impl ZorroDevice for Z3660 {
         self.pan_offset = 0;
         self.pan_width = 0;
         self.mode_set = false;
+        self.sprite_visible = false;
+        self.sprite_pix.clear();
+        self.sprite_colors = [0; 4];
     }
 
     fn kind(&self) -> &'static str {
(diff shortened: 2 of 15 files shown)


LinuxJedi (commented):
> [!NOTE]
> This is an AI code review performed by Claude Code (Fable 5), posted on behalf of @LinuxJedi.

# Review: z3660: emulate the Z3660 RTG board (Picasso96)

## Overview

Adds a new Zorro III device (`src/z3660.rs`, ~1360 lines) modelling the Z3660 accelerator's FPGA RTG core: autoconfig identity, the 32-bit register file at window offsets 0x100-0x7FF, 64 MB of backed board RAM, the GFXData mailbox blitter executor (fill/copy/invert/template/pattern/line/planar), palette capture, a 60 Hz vblank phase in emulated colour clocks, and a hardware-sprite mouse pointer composited at scanout. Presentation is hooked into both the window path (`render_rtg_frame_if_active`) and the CCP capture path (`render_frame` in `control/exec.rs`). Config plumbing (`[z3660] enabled`) follows the A2065 pattern exactly.

**Overall: high-quality, well-structured work that fits the codebase's conventions.** The tricky integration points were checked against the base tree and hold up. The blockers are stale documentation that contradicts the code, and two guest-controllable unbounded allocations.

## Verified as correct

- **Vblank timing under tick batching**: `is_idle()` defaults to `true`, but Zorro devices are ticked unconditionally in `tick_timed_devices`, and the deferred batch is flushed before every device-register read (`flush_timed_devices`), so `VBLANK_STATUS` advances correctly and `WaitVerticalSync` cannot hang. Batching is exact since `tick` is linear in cck.
- **Presentation buffer reshaping**: the RTG path resizes `present_fb` to `h * FB_WIDTH` (possibly smaller *or* larger than the native `FB_WIDTH * OUT_HEIGHT`). This is safe: `copy_present_frame`/`save_present_frame` scale arbitrary `src_rows` via `scaled_source_row`, the sync native path resizes per frame (`refresh_present_from_deinterlacer`), and the threaded path replaces the whole Vec.
- **Threaded-render race handling**: draining in-flight chipset results *before* the RTG compose is the right order -- a stale native frame cannot land on top of an RTG frame. At worst there is a one-frame native flash at the switch boundary, which self-corrects.
- **CCP capture consistency**: both `render_frame` consumers slice `fb[..FB_WIDTH * lines]`, matching the RTG buffer's exact size.
- **Save states**: `BoardDevice::Z3660` is correctly appended last (bincode variant-index rule); no existing struct changes shape, so no `STATE_VERSION` bump is needed. `MachineDescriptor` carries no board fields, consistent with existing boards.
- **Byte-order plumbing** (the trickiest part): the guest-order pixel stores, the 0x00RRGGBB palette/sprite-colour representation, and the RGBA-low-byte-R presentation output are all mutually consistent, including the sprite composite swizzle. The two-halfword register-landing rule (`off + size == aligned + 4`) correctly handles 68k split 32-bit writes.
- 128 MB is a valid Zorro III size; the register-file window shadows the first 2 KB of RAM harmlessly.

## Issues

### 1. Documentation contradicts the code (should fix before merge)

The user-facing docs describe an earlier iteration of the branch:

- `src/config.rs` -- both `Config::z3660` and `RawZ3660` doc comments say *"Bring-up stub: registers are logged, no display output yet."* The PR ships a full display pipeline.
- `copperline.example.toml` -- says *"planar blits, line draws, and surface ops are not implemented yet"*, but `OP_P2C`/`OP_P2D` and `OP_DRAWLINE` are implemented (the PR description confirms this). Only the ACC_OP surface ops are stubbed.
- `docs/guide/configuration.md` -- same stale list, plus the "RTG board stub" heading.
- `src/emulator.rs` -- the build comment and `info!` line also say "stub ... registers logged".

The module doc in `z3660.rs` is accurate; the rest should be reconciled to it. Given the project's docs-in-same-change rule, this is a merge blocker. Minor: consider a line in `docs/zorro.md` alongside the A2065 section.

### 2. Unbounded guest-controlled allocations (robustness)

Everything in the executor is carefully bounds-checked per byte **except** two allocations sized directly from mailbox fields:

- `OP_COPYRECT`'s overlap snapshot: `vec![0u8; x1 * y1 * bpp]` -- up to 65535 x 65535 x 4 = ~17 GB from one doorbell write. A buggy or hostile guest program aborts the emulator host.
- `OP_SPRITE_BITMAP`: `vec![0u8; w * h]` -- `w <= 16 * hires`, but `hires = y2 + 1` can be 65536, so this does not actually bound `w`; up to ~4 GB.

`rtg_frame` already clamps to 4096x4096; the executor should clamp rect/sprite dimensions similarly (real VRAM bounds them on hardware). This also caps the synchronous-blit stall time (a 65535x65535 FILLRECT is ~4.3 billion iterations on the doorbell write).

### 3. Logging volume and cost (style/perf)

- Every register read/write logs at `info!` (only VBLANK is demoted to `trace!`), and `log_gfxdata` logs every blit op at `info!` -- on a live Workbench desktop that is a firehose at the default level, for every user with the board enabled, not just driver developers.
- `log_gfxdata` builds three `Vec<u16>`s and does a dozen big-endian reads *before* the macro, so the cost is paid even when the level is filtered out -- on every blit.

Suggestion: keep `info!` for the one-time init probes (FW_VERSION, mode set, vidmode params), demote steady-state register traffic and per-blit mailbox dumps to `debug!`/`trace!`, and gate the `log_gfxdata` field extraction behind `log::log_enabled!`.

### 4. Minor points

- **Frozen display edge case**: `render_rtg_frame_if_active` returns `Some(false)` when `rtg_active()` is true but `rtg_frame` composes `None` (e.g. `MODE` written before `ORIG_RES`, or an invalid colormode). That suppresses the chipset render too, freezing the last frame indefinitely rather than falling back to native output. Returning `None` on compose failure would degrade more gracefully.
- **Main-thread compose cost**: the RTG path bypasses the render worker; at 1920x1200 that is ~2.3M per-pixel colormode matches plus a horizontal nearest scale per frame, synchronously. Fine for now, but worth a note next to the "pixel-perfect presentation" follow-up. Hoisting the colormode dispatch out of the pixel loop in `rtg_frame` would be a cheap win.
- **Save-state size**: the 64 MB `vram` rides the zlib stream. Mostly-zero VRAM compresses to almost nothing, but a busy 32-bit 1920x1200 desktop adds real megabytes and 64 MB of compressor work per save. Acceptable; worth knowing.
- `RECT_PATTERN` assumes `user0` is a power of two (`& (pat_rows - 1)`); if the firmware guarantees this, a comment saying so would help.

## Test coverage

Good breadth for a first pass: palette capture, activation gating, CLUT/565/32-bit scanout, pan, fillrect edges, *overlapping* copyrect, template JAM2, P2C and P2D decode, and the full sprite lifecycle. Gaps worth closing, since these regress silently:

- `OP_DRAWLINE` -- especially the COMPLEMENT XOR path the shell cursor depends on, and pattern rotation.
- `OP_RECT_PATTERN`, `OP_INVERTRECT`, `OP_COPYRECT_NOMASK` (the `pitch1`/`src`-surface variant).
- 15-bit (R5G5B5) scanout decode -- the only pixel format with no test.
- The split-write landing rule: two halfword writes high-first firing `reg_written` exactly once.
- `compose_rtg_present` scaling (both `w < FB_WIDTH` and `w > FB_WIDTH`).

## Security

The Zorro window is guest-facing attack surface; apart from the two allocations in issue 2, all VRAM accesses are bounds-checked and out-of-range accesses are dropped and logged. No host filesystem/network exposure. Determinism is preserved (all state advances on emulated cck), so headless runs and replay stay byte-identical.

## Verdict

Approve once issue 1 (stale docs) and issue 2 (allocation clamps) are addressed; issue 3 and the test gaps are strongly recommended but could be fast follow-ups.

---

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+                    "z3660: vidmode param {} = {value}",
+                    vmode_param_name(self.vidmode_param)
+                );
+            }
+            REG_BLITTER_DMA_OP => {
+                self.log_gfxdata("dma-op", dma_op_name(value), value);
+                if value == DMA_OP_PAN {
+                    // GFXData: offset[0] = framebuffer start relative to
+                    // MemoryBase, x[1] = row width in pixels.
+                    self.pan_offset = self.be32(GFXDATA_OFFSET);
```
LinuxJedi: **[P2] Apply PAN's x/y viewport offsets**

When `SetPanning` supplies a nonzero x or y offset, the driver places them in `x[0]`/`y[0]`, but this records only the raw framebuffer address. Consequently `rtg_frame` always scans from the bitmap's top-left instead of the selected viewport; include the signed offsets, pixel size, and row stride in the scanout base.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+        let scaled = (mode >> 12) & 0xF != 0;
+        let orig = self.reg_value(REG_ORIG_RES);
+        let (out_w, out_h) = (orig >> 16, orig & 0xFFFF);
+        let (mut w, mut h) = if scaled {
+            (out_w / 2, out_h / 2)
+        } else {
+            (out_w, out_h)
+        };
+        if self.pan_width != 0 {
+            w = self.pan_width;
```
LinuxJedi: **[P2] Keep PAN width as the row stride**

When a virtual bitmap is wider than the active display mode, PAN's `x[1]` is the bitmap stride rather than the visible width. Overriding `w` here causes the whole virtual row to be read and scaled into the window; keep the visible width derived from `ORIG_RES` and use `pan_width` only when calculating each row address.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+        const OP_INVERTRECT: u32 = 9;
+        const OP_SPRITE_XY: u32 = 11;
+        const MINTERM_NOTSRC: u8 = 3;
+        const MINTERM_SRC_IDX: u8 = 12;
+        const COMPLEMENT: u8 = 2;
+        const OP_SPRITE_COLOR: u32 = 12;
+        const OP_SPRITE_BITMAP: u32 = 13;
+        const MINTERM_SRC: u8 = 0xC0;
+        const JAM1: u8 = 0;
+        const INVERSVID: u8 = 8;
```
LinuxJedi: **[P2] Decode INVERSVID from bit 2**

Picasso96 and the Z3660 firmware encode `INVERSVID` as `0x04`. Defining it as `0x08` means the source pattern is never inverted for valid draw modes, so inverse text and patterned drawing render incorrectly.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+                let mut pattern = self.gfx_u16(0x20, 1) as u16;
+                if drawmode & INVERSVID != 0 {
+                    pattern ^= 0xFFFF;
+                }
+                let complement = drawmode & COMPLEMENT != 0;
+                let jam2 = !complement && drawmode & 1 != 0;
+                let steps = dx.abs().max(dy.abs());
+                let mut cur_bit = 0x8000u16;
+                for i in 0..=steps {
+                    let px = ax + if steps == 0 { 0 } else { dx * i / steps };
```
LinuxJedi: **[P2] Follow DRAWLINE's length and Bresenham stepping**

For non-axis-aligned lines, this interpolation does not match the board firmware's Bresenham rasterization; for example, `dx=2, dy=1` selects `(1,0)` instead of `(1,1)`. It also ignores `user[0]`, the requested `Line.Length`, so lines whose length differs from the largest delta draw the wrong number of pixels.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+                            let b = if at < self.vram.len() {
+                                self.vram[at] ^ inv
+                            } else {
+                                0
+                            };
+                            if b & bit != 0 {
+                                pen |= 1 << p;
+                            }
+                        }
+                        let at = dst + (dyr + y) * dpitch + (dxr + i) * bpp;
```
LinuxJedi: **[P2] Keep P2C destinations byte-addressed**

When a P2C operation follows an operation that left the mailbox color mode at 16 or 32 bits, `BlitPlanar2Chunky` does not rewrite that field. P2C output is always one byte per chunky pixel, but multiplying the destination x coordinate by the stale `bpp` writes sparse pixels at two- or four-byte intervals.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+            }
+            OP_SPRITE_BITMAP => {
+                // Amiga 2-plane sprite rows at offset[1]: `hires` words of
+                // plane 0 then plane 1 per row; pixel value = plane bits.
+                let (w, h) = (x1, y1);
+                let hires = y2 + 1;
+                // Row: `hires` words of plane 0, then plane 1 (the driver
+                // copies (w>>3)*2*hires bytes per row).
+                let row_bytes = (w / 8) * 2 * hires;
+                if w == 0 || h == 0 || w > 16 * hires {
```
LinuxJedi: **[P2] Honor doubled hardware sprites**

When Picasso96 enables `BIF_BIGSPRITE`, `SetSpriteImage` sets `x[2]` to request two-times scaling. This implementation ignores that flag and commonly rejects the resulting 32-pixel normal-resolution pointer as wider than `16 * hires`, clearing the sprite instead of doubling its decoded pixels in both axes.

src/zorro_device.rs
```diff
@@ -313,6 +313,7 @@ pub enum BoardDevice {
     // Appended last: bincode encodes variants by index, so inserting
     // anywhere else would break save states holding the boards above.
     Filesys(crate::filesys::FilesysBoard),
+    Z3660(crate::z3660::Z3660),
```
LinuxJedi, line 337: **[P2] Bump the save-state format for the new enum variant**

A state containing this serialized `BoardDevice` variant cannot be opened by another build that still advertises state version 31: the version check succeeds and bincode fails later on the unknown variant. Adding a serialized enum variant is a format-shape change, so `savestate::STATE_VERSION` must be bumped as required by the repository's save-state contract.

src/z3660.rs
```diff
@@ -0,0 +1,1364 @@
...
+            return None;
+        }
+        Some((u16::from(self.vram[at]) << 8) | u16::from(self.vram[at + 1]))
+    }
+
+    fn tick(&mut self, cck: u32, _host: &mut DeviceHost) {
+        self.frame_phase = (self.frame_phase + cck) % FRAME_CCK;
+    }
+
+    fn reset(&mut self) {
```
LinuxJedi, line 927: **[P2] Add cold-reset handling for VRAM**

After guest writes modify Z3660 VRAM, a host power-off/power-on calls the same device reset path but never clears `vram`. This violates `Bus::power_on_reset`'s zeroed-RAM contract and lets stale framebuffer data survive a cold boot, so the board needs cold-reset-specific VRAM clearing while retaining it across warm resets if desired.

LinuxJedi (changes requested):
> [!NOTE]
> This is an AI code review performed by OpenAI Codex, posted on behalf of @LinuxJedi.

I found eight correctness issues not already covered by Claude's review. They affect panning, inverse drawing, line rasterization, planar conversion, doubled sprites, save-state compatibility, and cold-reset behavior. Requesting changes until these are addressed.

src/z3660.rs
codewiz: Fixed in 6a35f7e: corrected the constant to 0x04 (bit 2), matching the firmware rtg/gfx.h header (JAM bit 0, COMPLEMENT bit 1, INVERSVID bit 2).

src/z3660.rs
codewiz: Fixed in 8552fcf. This was in fact what made the pointer vanish once BIGSPRITE was enabled: the decoder now handles wide and doubled sprites (plane-0-then-plane-1 row layout per the firmware, half-size source scaled 2x2 for a doubled sprite), clamped to the 32x48 buffer, verified against the OS 3.2 32x48 pointer.

src/zorro_device.rs
codewiz: Fixed in 1c61746: bumped STATE_VERSION to 32 with a changelog entry.

src/z3660.rs
codewiz: Confirmed on both counts (for dx=2, dy=1 the interpolation lands on (1,0) not (1,1), and it ignores user[0] = Line.Length). Porting the firmware Bresenham stepping loop in a follow-up commit.

src/z3660.rs
codewiz: Confirmed: P2C output is always one byte per chunky pixel, and BlitPlanar2Chunky does not reset the mailbox colormode, so a stale 16/32-bit value mis-strides the destination. Will force a byte-addressed destination for P2C.

src/z3660.rs
codewiz: Agreed: reset() clears the registers but not VRAM, and it is shared by warm and cold resets. Will thread a cold/warm flag so a cold power-on zeroes VRAM while a warm reset preserves it.

src/z3660.rs
codewiz: In every capture I have from a live Workbench session, SetPanning arrives with x=y=0, so the current scanout from the raw framebuffer address matches. I would rather reproduce a nonzero pan offset before adding the viewport math than write code I cannot exercise. Do you have a setup that scrolls a virtual or oversized screen? Happy to add it once I can confirm the path.

src/z3660.rs
codewiz: Same as the x/y-offset thread: in my captures x[1] equals the visible width, so overriding w is a no-op today, but you are right for a bitmap wider than the mode. I will pair this with the viewport-offset fix once I can reproduce a wider-than-visible bitmap.

src/z3660.rs
codewiz: On reflection, not changing this. The current interpolation already draws a valid connected line from the endpoint along the delta; it differs from the firmware Bresenham by at most one pixel row on a diagonal, which the P96 DrawLine contract does not pin down. Matching the exact firmware rasterization would be bug-for-bug parity, not spec conformance, for +100 lines that are not simpler, faster, or more correct per P96. Line.Length equals the major-axis span on the P96 path we see, so it is honored implicitly. Happy to revisit if a real title draws visibly wrong lines.

src/z3660.rs
codewiz: Fixed in e76687c: P2C now uses a fixed 1-byte destination stride, and the regression test sets a stale 16-bit colormode on the P2C op to guard it.

src/z3660.rs
codewiz: Fixed in 8c5400c. I went with the simpler option rather than a cold/warm flag: reset() now clears VRAM unconditionally. A warm reset re-inits the board and repaints, so clearing there too is harmless, and it satisfies the zeroed-RAM contract on a cold boot without threading a new signal through the device trait.

src/z3660.rs
codewiz: Correcting my earlier reply on this thread: the P96 DrawLine specification does pin the rasterization down -- it requires Bresenham, and states that Line.Length is authoritative over the raw dX/dY, which is what lets a clipped segment draw the right pixel count. You were right on both counts and my pushback was wrong. Fixed in df8c96c: the op now steps with Bresenham and honours Length (mailbox user[0]), with tests for the dx=2, dy=1 diagonal and a Length-clipped span. Verified end to end with a small P96 harness that draws 24-ray stars and diffs the readback -- 4804 pixels change, all on the diagonal rays, with the axis-aligned rays untouched as expected.

src/z3660.rs
codewiz: Fixed in 6641641.

You were right, and reproducing it first was worth it: a 1024x768 screen on an 800x600 mode pans as you push the pointer to the edges, which exercises both this and the stride thread.

The offsets now fold into the scanout base the way the board does it:

    pan = offset[0] + (x[0] + y[0] * x[1]) * bpp

One deliberate deviation. The firmware derives this by shifting the offsets by the GFXData colormode (`dma_rtg.c` OP_PAN), which equals the pixel size for 8/16/32-bit but strides 8 bytes per pixel for colormode 3 (15-bit). We scan by the pixel size instead, since P96 semantics are what a driver gets checked against.

I also tried shifting the sprite by the same offsets, since `do_update_hw_sprite_pos` does `x - vs.sprite_x_offset + 1` (`video.c:1441`). That turned out to be wrong: `SetSpritePosition` already sends `x - b->XOffset` (`gfx.c:2355`), so the firmware subtracts twice and the pointer drifts away from where clicks land. Manual testing caught it. We do not reproduce that one.

Covered by `pan_viewport_offsets_move_the_scanout_origin`.

src/z3660.rs
codewiz: Fixed in 6641641, paired with the viewport-offset thread as promised.

The visible size now comes from `ORIG_RES` and `pan_width` is used only as the row stride, matching `video.c:806`, where the firmware reprograms the VDMA stride from `framebuffer_pan_width` precisely when it differs from `hsize / hdiv`.

Verified both ways: a 1024x768 screen on an 800x600 mode renders correctly while panning, and an unpanned Workbench still scans out with no row shear, which is what a wrong stride would show first.

Covered by `pan_width_is_a_stride_not_a_visible_width`.

src/z3660.rs
```diff
@@ -0,0 +1,1436 @@
...
+        }
+        let bpp: u32 = match colormode {
+            COLORMODE_8BIT => 1,
+            COLORMODE_16BIT565 | COLORMODE_15BIT => 2,
+            COLORMODE_32BIT => 4,
+            _ => return None,
+        };
+        let pitch = (w * bpp) as usize;
+        let base = (VRAM_OFFSET + self.pan_offset) as usize;
+        out.clear();
```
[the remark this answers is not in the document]
codewiz: Addressed in 6641641 while reworking this path for the two PAN threads: the base now uses `checked_add` and returns `None` on overflow, so the presenter falls back rather than scanning an unrelated region.

The pan offset itself is also computed in `i64` now, since it folds in signed viewport offsets, and is clamped to a non-negative `u32` before it is stored.

src/video/present_common.rs
```diff
@@ -18,6 +18,32 @@ pub const fn rgba(r: u32, g: u32, b: u32) -> u32 {
...
+    let (w, h) = bus.rtg_frame(scratch)?;
+    let (w, h) = (w as usize, h as usize);
+    out.clear();
+    out.resize(h * FB_WIDTH, 0xFF00_0000);
+    for y in 0..h {
+        let src = &scratch[y * w..(y + 1) * w];
+        let dst = &mut out[y * FB_WIDTH..(y + 1) * FB_WIDTH];
+        for (x, px) in dst.iter_mut().enumerate() {
+            *px = src[x * w / FB_WIDTH];
+        }
```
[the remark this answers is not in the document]
codewiz: Fixed in 4b7075a, though with centre sampling rather than endpoint mapping: `(2 * x + 1) * w / (2 * FB_WIDTH)`. That reaches the last source column, which was the defect, and stays unbiased across the row, whereas mapping the endpoints stretches the sampling grid to make the extremes line up.

Scope, for the record: the effect is confined to screenshots and CCP captures, and only when the board frame is wider than FB_WIDTH (716), so 800x600 and up. The window presents the native RTG frame directly and never went through this path.

src/z3660.rs
```diff
@@ -0,0 +1,1436 @@
...
+        const MINTERM_SRC: u8 = 0xC0;
+        const JAM1: u8 = 0;
+        // Drawmode bits (firmware rtg/gfx.h): JAM1/JAM2 in bit 0, COMPLEMENT
+        // in bit 1, INVERSVID in bit 2.
+        const INVERSVID: u8 = 4;
+
+        let g = GFXDATA_OFFSET;
+        let dst = VRAM_OFFSET as usize + self.be32(g) as usize;
+        let src = VRAM_OFFSET as usize + self.be32(g + 4) as usize;
+        let rgb0 = self.be32(g + 8);
```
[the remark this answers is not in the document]
codewiz: Fixed in 3a9f373. The bases are now computed in u64 and saturated back to usize, so an out-of-range base stays out of range and every access below fails its existing bounds check, which is also what the firmware does when its writes land in scratch.

Saturating rather than bailing out is deliberate: plain COPYRECT never reads `src`, so returning early on a stale or bogus source offset would drop blits that are perfectly valid.

Worth noting the concrete case, since it is not hypothetical on 32-bit: with usize at 32 bits, offset 0xFFFF_FFF0 wrapped to 0x1FFFF0, which is inside the GFXData mailbox rather than merely the wrong part of VRAM. `blit_base_near_the_address_space_top_does_not_wrap` guards it, though it only discriminates where usize is 32 bits -- on a 64-bit host the sum is simply far out of range and the op drops either way. I checked `crates/copperline-web` still builds for wasm32.

src/z3660.rs
```diff
@@ -0,0 +1,1436 @@
...
+                    (src, pitch1 * 4)
+                };
+                if op == OP_COPYRECT_NOMASK && minterm != MINTERM_SRC {
+                    log::debug!("z3660: copyrect minterm {minterm:#04x} treated as SRC");
+                }
+                let apply_mask = op == OP_COPYRECT && bpp == 1 && mask != 0xFF;
+                // Snapshot the source rect first: rects may overlap.
+                let mut rect = vec![0u8; x1 * y1 * bpp];
+                for y in 0..y1 {
+                    let srow = sbase + (y2 + y) * spitch + x2 * bpp;
```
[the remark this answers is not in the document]
codewiz: Fixed in 3a9f373, taking the in-place option you suggested rather than a byte cap.

The rect snapshot is gone: the op now buffers one row (x1 * bpp, so 16 KiB at the clamp) and walks rows away from the destination, choosing the direction from whether the destination starts after the source. That keeps overlapping blits correct -- the reason the snapshot existed -- without the allocation scaling with the guest rect at all.

I first tried simply bounding the rect by VRAM, which turned out to be nearly worthless: VRAM is 64 MiB and the clamped worst case is 64 MiB, so the pathological blit still allocated the lot. `copyrect_scrolls_down_without_smearing` covers the direction, and fails with a top-first walk.

LinuxJedi (approved):
Fantastic stuff. Many thanks for adding this.

PR #218, opened 2026-07-18
serial: browser WebSocket bridge and native tcp-connect dial-out

LinuxJedi:
## What

Route A of the Retro32 BBS request: run a telnet BBS session from the browser build, with the guest side being a plain terminal program on serial.device.

- **Wasm serial transport**: `ChannelSerialSink`/`ChannelSerialHandle` in `serial.rs` (thread-free, works on wasm32). `WebEmu` exports `serial_send(bytes)`, `serial_take()`, and `serial_input_backlog()`; the channel sink replaces the stdout sink in browser machines and survives resets/ROM swaps. Guest output is bounded (drop-oldest + counter) so a non-draining page cannot balloon memory; input is unbounded by design (the UART consumes at emulated baud) with the backlog exposed for flow control.
- **Telnet NVT layer** (`www/serial-telnet.js`): answers option negotiation (ECHO, SGA, BINARY for ZModem, TERMINAL-TYPE replying "ANSI"), refuses everything else, handles IAC escaping and CR NUL both ways, and survives negotiation split across WebSocket frames.
- **Page glue** (`www/try.js`): a shell providing `#serial-url`/`#serial-connect` (+ optional `#serial-status`/`#serial-raw`) gets the whole connect/disconnect/telnet/flow-control loop; pages without the elements are untouched. The wasm-demo workflow ships the new file.
- **Native parity**: `[serial] mode = "tcp-connect"` + `connect = "host:port"`, or `--serial-connect HOST:PORT` (implies the mode) — the outbound counterpart of the existing tcp listen mode, for dialling a BBS directly from the desktop build.

Docs updated in the same change: `docs/guide/configuration.md` (new mode), `docs/guide/browser.md` (serial bridge section: API, page contract, websockify-style gateway shape), `copperline.example.toml`.

## Testing

- `cargo test` green (1617 unit tests, including new channel-sink round-trip/bounding tests, tcp-connect socket round-trip, config/CLI/launcher round-trips); `cargo clippy` and `cargo fmt --check` clean; web crate builds for `wasm32-unknown-unknown`.
- Telnet layer exercised under node: 14 behavioral checks (negotiation, refusals, TTYPE subnegotiation, IAC/CR escaping both directions, binary mode, chunk-split sequences).
- Native smoke test: `--serial-connect` against a local listener connects, logs, and exits cleanly.
- Browser end-to-end: local copy of the /try page with this bundle + a scripted WebSocket "fake BBS"; booted AROS in Chrome, clicked Connect, and verified the negotiation on the server side (`IAC DO ECHO`, `DO SGA`, `WILL TTYPE`, then `TTYPE IS ANSI`), banner bytes draining into the guest UART, and a clean disconnect on both ends.

### DIFF
--- a/.github/workflows/wasm-demo.yml
+++ b/.github/workflows/wasm-demo.yml
@@ -9,8 +9,9 @@ name: Browser demo
 # The page shell (try/index.html) is hand-written in the website repository
 # and left alone. This workflow replaces the parts that must change together
 # with the emulator: the wasm-bindgen bundle (try/pkg), the JS glue that
-# drives the WebEmu API (try/try.js, try/audio-worklet.js, sourced from
-# crates/copperline-web/www), and the AROS ROMs (try/aros).
+# drives the WebEmu API (try/try.js, try/serial-telnet.js,
+# try/audio-worklet.js, sourced from crates/copperline-web/www), and the
+# AROS ROMs (try/aros).
 #
 # Pushing to the website repository needs the SITE_DEPLOY_KEY secret: the
 # private half of an SSH deploy key whose public half is installed with
@@ -63,6 +64,7 @@ jobs:
           cp crates/copperline-web/pkg/copperline_web.js \
              crates/copperline-web/pkg/copperline_web_bg.wasm site/try/pkg/
           cp crates/copperline-web/www/try.js \
+             crates/copperline-web/www/serial-telnet.js \
              crates/copperline-web/www/audio-worklet.js site/try/
           cp assets/aros/aros-amiga-m68k-rom.bin \
              assets/aros/aros-amiga-m68k-ext.bin \
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -382,13 +382,18 @@ floppy_sounds_volume = 100
 #   "tcp"     serial in/out is bridged to a host TCP port, like UAE's "TCP:"
 #             device. listen sets the bind address (default 127.0.0.1:1234);
 #             connect with e.g. `nc`, `socat`, or a raw-mode telnet client.
+#   "tcp-connect"
+#             the outbound counterpart of "tcp": dial the remote named by
+#             connect (required, "host:port") at startup - a telnet BBS, a
+#             tcpser modem bridge, any TCP byte service.
 #   "pty"     serial in/out is bridged to a host pseudo-terminal (Unix only).
 #             The slave path (/dev/pts/N) is logged at startup; attach a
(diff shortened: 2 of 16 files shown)


LinuxJedi:
Follow-up commit: exercising the bridge end-to-end with a real guest terminal (a bootable serial-console disk driving Paula's UART at 19200) exposed two emulator-side RX gaps that dropped bytes real hardware would have delivered -- host bytes queued in the sink did not bound the serial event horizon while the receive shift was idle, and the STOP-state idle fast-forward could nap through a burst arriving on the host's clock. Both fixed in 75af2d0 with a unit test; verified byte-perfect against a scripted TCP BBS on KS2.05 and KS3.1 across repeated runs. Default machines (stdout serial sink) are unaffected.

LinuxJedi:
Two more commits:

- **93d9607** -- the intended AROS-ROM rebuild turned out to be unnecessary: the real blocker was another Paula modelling gap. Clearing INTREQ RBF erased the received word, but on real hardware the receive buffer is a physical latch that survives the ack (only the flag and OVRUN drop). AROS's level-5 dispatcher acks INTREQ *before* running the RBF handler chain (`arch/m68k-amiga/kernel/amiga_irq.c`, `PAULA_IRQ_ACK`), so under the old model every received byte was destroyed before any handler saw it. With the latch modelled correctly, guest serial input works on the bundled AROS ROM out of the box -- verified end-to-end with a bare-metal terminal boot disk on AROS, KS2.05, and KS3.1 (overrun is now gated on the RBF flag, with a regression test for the ack-first pattern).
- **38d1562** -- both Copilot review points addressed: tcp-connect is offered in the launcher cycle only when the loaded config carries a dial-out address (no launcher editor exists for it), and the telnet layer clears its CR state when an IAC sequence starts so an unrelated NUL after a negotiation command is no longer swallowed.

LinuxJedi:
One more emulator fix surfaced while polishing the terminal disk's AROS rendering (e3aa80a): the serial tick's interrupt bits were OR'd into INTREQ without the slice preemption every other interrupt source gets, so a running CPU could stay blind to RBF past a character time -- a deterministic first-character loss on each burst under AROS, whose scheduler busy-idles (and masks interrupts in ~1 ms stretches) instead of STOPping. Serial completions now latch through `latch_interrupt_sources` and preempt the slice, with a bus regression test. With this plus disk-side changes (SA_Colors palette, per-ROM baud, a CSI reassembly/split shim for the AROS console dialect), the terminal now renders full ANSI colour on black, byte-perfect, on the bundled AROS ROM and KS2.05/3.1 alike.

PR #279, opened 2026-07-25
Allow A3000/A4000 mb ram in plausible_ptr()

codewiz:
First PR I do entirely on my own in a while :-)

Let's see if it's correct.

### DIFF
--- a/src/amigaos.rs
+++ b/src/amigaos.rs
@@ -103,8 +103,10 @@ pub fn task_state_name(state: u8) -> &'static str {
 /// moves ExecBase to fast RAM, which may be a Z3 board).
 fn plausible_ptr(addr: u32) -> bool {
     addr & 1 == 0
-        && ((0x100..0x00A0_0000).contains(&addr)          // chip + Z2 fast
-            || (0x00C0_0000..0x00D8_0000).contains(&addr) // slow
+        && ((0x100..0x00A0_0000).contains(&addr)           // chip + Z2 fast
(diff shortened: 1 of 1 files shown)


src/amigaos.rs
```diff
@@ -103,8 +103,9 @@ pub fn task_state_name(state: u8) -> &'static str {
 /// moves ExecBase to fast RAM, which may be a Z3 board).
 fn plausible_ptr(addr: u32) -> bool {
     addr & 1 == 0
-        && ((0x100..0x00A0_0000).contains(&addr)          // chip + Z2 fast
-            || (0x00C0_0000..0x00D8_0000).contains(&addr) // slow
+        && ((0x100..0x00A0_0000).contains(&addr)           // chip + Z2 fast
+            || (0x00C0_0000..0x00D8_0000).contains(&addr)  // slow
+            || (0x0700_0000..0x0800_0000).contains(&addr)  // Ramsey fast
```
LinuxJedi: Should also have 04000000 - 06FFFFFF too, which is the extended A4000 range as well as 08000000 - 0FFFFFFF for non-Zorro accelerator area.

LinuxJedi (changes requested):

src/amigaos.rs
codewiz: Done. I left the adjacent intervals separate for clarity, hopefully the optimizers will merge them.

LinuxJedi (approved):

Issue #278, opened 2026-07-25
MINOR Graphics Issues


Gordon1972:
Hi there @Copperline,
just wanted to bring up some MINOR graphic glitches to mind that are present in the current build 0.13.0 (Windows):.
I'm using the default A500 configuration with .adf files loaded:

The Great Giana Sisters (Lines in the Logo):
<img width="1217" height="279" alt="Image" src="https://github.com/user-attachments/assets/d423c766-bcba-46ca-825c-efb4902647a9" />

Hybris (Dots under "Lives"):
<img width="944" height="931" alt="Image" src="https://github.com/user-attachments/assets/64577d8e-b17d-4ddd-9f60-f588027373ec" />
Maybe it's not the top priority right now, just a reminder.

**P.S.: We'd really appreciate it if you'd implement some of the keyboard-shortcuts for example Fullscreen or the Status Bar
           to the .toml configuration file and to the command-line (if possible).
           Maybe a Switch to turn on/off/auto the Power LED (Audio Filter) as well? That would be very nice.**

For Example (.toml-file):

[video]
fullscreen = on/off
statusbar = on/off

[audio]
filter = on/off/auto

Thanks for reading and keep up the good work on Copperline!

LinuxJedi:
Thanks @Gordon1972, we'll look into this ASAP. And the .toml options are a good idea too.

hobbo91:
@LinuxJedi happy to work on the audio filter overrides today, was thinking about this a while back but never go around to it.

LinuxJedi:
@hobbo91 go ahead. I’m currently working on the main bug.

LinuxJedi:
PR 280 fixes most of the two bugs, but there is one minor issue with The Great Giana Sisters still to fix...

Edit: scratch that, I was using the wrong build. Both bugs should be fixed now. Leaving open for @hobbo91's work.

@Gordon1972 Windows build from main should be available here: https://github.com/CopperlineHQ/Copperline/actions/runs/30159174589/artifacts/8619802354

Gordon1972:
> PR 280 fixes most of the two bugs, but there is one minor issue with The Great Giana Sisters still to fix...
> 
> Edit: scratch that, I was using the wrong build. Both bugs should be fixed now. Leaving open for [@hobbo91](https://github.com/hobbo91)'s work.
> 
[...]

PR #281, opened 2026-07-25
Log Denise's hardware sprite latch view in the frame-state dump

LinuxJedi:
Small debugger addition that fell out of the #278 investigation.

`COPPERLINE_DBG_FRAMESTATE` printed only the CPU/Copper write shadow of the
sprite registers:

```
  sprpos=[FE00, FE00, ...] sprctl=[FF00, FF00, ...] sprarmed=[false, ...]
```

Denise keeps a second, hardware-true view of those same registers (`spr_hw_*`)
which sprite DMA fetches write through as well as CPU/Copper writes, and that
view -- not the shadow -- is what drives the DMA-idle latched redisplay. So a
frame with a stale sprite column but no sprite register writes at all looked
completely clean in the dump. Now both views print:

```
  sprpos=[FE00, FE00, ...] sprctl=[FF00, FF00, ...] sprarmed=[false, ...]
  spr_hw_pos=[2B3F, FE00, ...] spr_hw_ctl=[3B01, FF00, ...] spr_hw_data=[0080, 0000, ...]
  spr_hw_datb=[0140, 0000, ...] spr_hw_armed=[true, false, ...]
```

That is the real output from the state I was handed while chasing the residual
Giana Sisters column: the program had parked all eight channels with the
`SPRxPOS=$FE00 / SPRxCTL=$FF00` idiom (a CTL write, which disarms), sprites 1-7
show exactly that in both views, and sprite 0 alone still held its DMA-fetched
descriptor words with the latch armed. One line, whole diagnosis -- worth having
next time a sprite appears out of nowhere.

Diagnostic-only: the knob is off by default, so the added line costs nothing in
a normal run. Documented in `docs/debugger/headless.md` alongside the knob,
including what a disagreement between the two views means.

### DIFF
--- a/docs/debugger/headless.md
+++ b/docs/debugger/headless.md
@@ -131,16 +131,17 @@ cannot change at runtime (see [](../internals/architecture)).
   `COPPERLINE_DBG_EXPORT_PLANES_DIR=DIR` sets the output directory.
 
 `COPPERLINE_DBG_FRAMESTATE=1`
-: Log the per-frame display state (one summary line per rendered frame)
-  inside the `AFTER`/`UNTIL` window, followed by the frame geometry, the
-  palette, a captured-sprite-DMA summary, and *both* of Denise's sprite
-  register views: the CPU/Copper write shadow (`sprpos`/`sprctl`/`sprarmed`)
-  and the hardware-true latch view (`hw_sprpos`/`hw_sprctl`/`hw_sprdata`/
-  `hw_sprdatb`/`hw_sprarmed`), which sprite DMA fetches write through as
-  well. The two disagreeing on a channel is the signature of a stale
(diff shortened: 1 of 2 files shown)


LinuxJedi:
Both taken, thanks:

- The hardware-view labels are now the field names themselves (`spr_hw_pos`/`spr_hw_ctl`/`spr_hw_data`/`spr_hw_datb`/`spr_hw_armed`), so a line grepped out of a log leads straight back to the struct it came from. Comment, output and fields all agree now, and the docs and the example in the PR body were updated to match.
- The `COPPERLINE_DBG_FRAMESTATE` entry now describes the block it actually emits (framestate summary line, geometry, palette, sprite-DMA summary, both register views) instead of "one summary line per rendered frame". That wording was already stale before this change; it is accurate now.

PR #292, opened 2026-07-26
launcher: add a Boot Priority sub-page for hard-disk drives

hobbo91:
Closes the "Not covered" item from the hard-disk boot-priority work in #286: the launcher could load and re-save each drive's synthesized-RDB `bootpri` but had no editor for it, so it could only be set in the TOML. This adds a **Boot Priority** sub-page to the *Hard Disk* tab.

## What it does

The Hard Disk tab links two sub-pages for **Boot Priority** and **Host Mounts**. The Boot Priority page has one row per hard-disk drive (IDE master/slave, SCSI units 0–6) under **Drive** / **Priority** / **Status** columns, each setting the `de_BootPri` written into the partition Copperline synthesizes in front of a bare hardfile (no effect on an image carrying its own RDB).

- **Priority**: `<`/`>` nudge by one; the value box is also a text field — type any priority in -128..127, Enter to commit.
- **Status**: a **Bootable** tick box, ticked by default. Clearing it greys that row's priority and writes the `-128` "disabled" sentinel. Priority and the Bootable flag are stored separately, so clearing the box keeps the number for the greyed display and re-ticking restores it within the session (the config can only hold one, so a save of a cleared box still writes `-128`).
- **Cascade defaults**: drives added in the launcher with no priority of their own don't tie — the first is `0` and each later one drops below the floppies: `-35`, `-40`, `-45`… A drive already carrying a priority keeps it, and a config that only names a device stays at `0`, so existing files are unchanged.
- Empty slots and CD images are greyed ("no drive" / "CD-ROM"), and a greyed **Help** line states the valid range and the floppy-drive priorities the cascade sorts around.

## Round-trip

GUI and TOML stay interchangeable: `bootpri = 5` loads as `5`; `-128` loads as a cleared Bootable box; a drive left at `0` with no cascade writes no key. The state and config plumbing already existed (values were carried through unedited) — this PR is the editor, its sub-page, and the two sub-page links.

## Notes

Confined to `src/video/{launcher,ui,window}.rs` (all frontend-gated — not built for the web crate) plus docs. No config/CLI/serde changes. `build --release`, `clippy --all-targets --all-features`, `fmt --check`, and the full test suite are clean; new tests cover the round-trip, the `-128`↔cleared-box mapping, cascade seeding, step clamping, and typed-entry parsing.

Screenshot 2026-07-26 at 01 45 01
Screenshot 2026-07-26 at 01 45 24


LinuxJedi (approved):
Looks great! Many thanks for adding that layout.

PR #294, opened 2026-07-26
cpu: charge the 020 pipeline refill on a taken branch

### LINKED ISSUE #285: A1200 is still slightly too fast
Cubic Dream Demo by TRSI crashes at the beginning with the default A1200 config. Copperline version is 0.13.0
If I reduce the clock speed to ~10MHz (3x CCK) it works fine.
This demo has a micro benchmark built in.
In FS-UAE it takes 5.31s to decrunch.
In Copperline: 5.08s
Small difference - huge impact. :-)
Interestingly recent *uae also has this issue, so it is not a good reference...

Link to the demo:
https://janeway.exotica.org.uk/release.php?id=3406

LinuxJedi:
Fixes #285.

## What goes wrong

Not a general "too fast" — Cubic Dream is racing itself, with 33 frames of margin.

- The demo's level-3 (VERTB) autovector points at `$7888C`. The picture part is a state machine at `$78BC2`; entering state 3 starts a **33-frame countdown**, after which the VERTB handler at `$78A80` installs the next part's handler (`$6C` -> `$7CD96`).
- Meanwhile the trackloader (wait loop `$1030`, MFM store at `$108C`) streams the next part over `$78xxx` — *including the live handler*.
- Copperline reached the overwrite at **state3+28 frames**, five frames before the handover. The next VERTB executed loaded data (`8240 A643`) -> Line-A -> Guru `8000 000A` at t=39.86s.

## Root cause

The first commit adds an **FS-UAE A1200 reference column** for `timing-test/`. The repo only ever had the 7 MHz A500+ one, and vAmiga cannot arbitrate an A1200 (OCS/ECS, A500/A1000/A2000 only), so there was no cross-check at the clock every AGA demo actually runs at.

It isolated the error cleanly: rows 4, 5, 7, 14, 28, 29, 30 were each exactly **two clocks per loop iteration** fast, and only in the branch — `move`, `shift`, `mul` and every chip-bus/blitter/beam row already agreed.

`timing_020.rs` bills the MC68020UM section-8.2 entry for the branch in isolation, but the manual accounts for the post-flush pipeline refill in the *following* instruction's head, which a per-instruction model with no overlap stage has nowhere to put. `TAKEN_BRANCH_REFILL` charges it where the flush happens. A `dbra` loop goes 6 -> 8 clocks per iteration; the demo's margin goes 28 -> 42 frames and it plays through.

Second, independent fix: a 020 custom-register read now costs one colour clock beyond a chip-RAM read. Chip RAM answers from Agnus' DRAM controller with the row already open; a register read crosses to Agnus/Denise/Paula and back over the 16-bit chipset bus first. A 68000's four-clock bus cycle is longer than either path so it cannot see the difference. This sets the rate of every VHPOSR/INTREQR poll loop — rows 16, 17 and 21 were fitting 27% more iterations per frame.

17 of the 27 comparable rows now match the reference exactly.

## Still open

Chip **writes** are billed a whole colour clock where a real 020 posts inside its shorter bus cycle: rows 3/10/12 ~10% slow, row 18 ~29% slow under bitplane DMA. That error used to partially cancel the branch one — the demo's own decrunch benchmark read 5.08s before and reads 5.54s now, against FS-UAE's 5.31. Fixing it needs the sub-colour-clock write-posting work. Row 31 (DIV during an active display) is ~27% fast; `MULU.W #imm` is +2. All recorded in `timing-test/README.md`.

Separately worth knowing: `[cpu] clock_mhz` quantises to integer colour-clock multiples, so 12.5–16 MHz all mean 4x CCK. @sonnenscheinchen's "~10 MHz" was 3x CCK = 10.64, and there is no way to sweep CPU speed finely through config today.

## Verification

- 1827 root tests + 94 `crates/m68k` tests pass, clippy and fmt clean. (`cargo test` at the root **skips** `crates/m68k` — it is a path dep, not a workspace member. Its DBcc test only failed under `cargo test --lib` inside the crate.)
- 23 golden probe renders unchanged — no re-bless needed.
- Roots II AGA, Zool and the A1200 Kickstart boot render pixel-identical; the CD32 intro shifts animation phase with no blank gap (250-frame luminance scan clean).
- Cubic Dream runs clean past 110s through the later parts.

https://claude.ai/code/session_01BRCdzExVHg1HVUXx7K9AoU

### DIFF
--- a/crates/m68k/src/core/timing_020.rs
+++ b/crates/m68k/src/core/timing_020.rs
@@ -631,16 +631,16 @@ fn group_5(cpu: &CpuCore, op: u16, cached: bool) -> Option<i32> {
     }
     if size_bits == 3 && mode == 1 {
         let condition = ((op >> 8) & 0xF) as u8;
+        // Both exits fall through with the pipeline intact; only the looping
+        // arm branches, so only it pays the refill.
         let case = if cpu.test_condition(condition) {
             Case::new(6, 7)
         } else if cpu.d(reg as usize) as u16 == 0xFFFF {
             Case::new(10, 10)
         } else {
-            Case::new(6, 9)
+            Case::new(6 + TAKEN_BRANCH_REFILL, 9 + TAKEN_BRANCH_REFILL)
         };
-        let taken = !cpu.test_condition(condition) && cpu.d(reg as usize) as u16 != 0xFFFF;
-        let refill = if taken { TAKEN_BRANCH_REFILL } else { 0 };
-        return Some(case.pick(cached) + refill);
+        return Some(case.pick(cached));
     }
     if size_bits == 3 {
         if mode == 0 {
--- a/crates/m68k/src/core/timing_020.rs
+++ b/crates/m68k/src/core/timing_020.rs
@@ -18,9 +18,23 @@
 //! brief form after the instruction has retired. Indexed entries therefore
 //! use the manual's brief-form row; the host still bills every extension and
 //! indirect operand access that actually occurred.
+//!
+//! A taken branch additionally pays `TAKEN_BRANCH_REFILL`. The section-8.2
+//! entries are written for an instruction whose successor is already in the
+//! three-stage pipeline; a taken branch invalidates the decode and execute
+//! stages, so the target instruction cannot begin until the pipe refills from
+//! the cache. The manual accounts for that in the *following* instruction's
+//! head, which a per-instruction model with no overlap stage cannot express,
+//! so the cost is charged where the flush happens. Without it a `dbra` loop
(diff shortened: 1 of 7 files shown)


LinuxJedi:
Both Copilot comments applied in 2f2cd0c0.

- **DBcc double evaluation**: real point, and it is a per-instruction hot path. Folded the refill into the looping arm's `Case` instead of re-testing after the fact, so `test_condition` and the expiry check are each evaluated once. No behaviour change — the A1200 reference column and the Cubic Dream run are identical before and after.
- **`/dev/null`**: now `os.devnull`.

PR #304, opened 2026-07-26
0.14 debugging: automation primitives and the WinUAE-inspired debugger set

LinuxJedi:
The debugging features from `ROADMAP-0.14.md`: Tier 1's automation
primitives and the WinUAE-inspired debugger set from Axis 4. Nine
commits, one per feature, each with tests and docs.

## Tier 1 -- "the machine you can drive"

**`capture.region_digest {x?, y?, w, h}`** hashes one rectangle of the
frame, so a script can ask "is this button highlighted?" and get a
stable answer while the rest of the display moves. A rectangle that does
not fit the current frame is `-32602` rather than a silent clamp: frame
geometry moves with the beam standard and the canvas scale, so stale
coordinates need to be heard about.

**`run_until {stable_frames: N, max_frames?, x?, y?, w?, h?}`** runs
until N consecutive frames hash identically -- how to wait for a GUI to
finish drawing without guessing an emulated-seconds delay. The optional
region is what makes it usable on a live Workbench: a blinking cursor
never lets a whole frame settle, but the dialog being waited on does.
Both server modes share one `StableWatch`, and the settle/give-up state
machine is split from the rendering so it can be tested on synthetic
digest sequences.

**`input.mouse_to` and `--mouse-to-after SECS X Y [PORT]`** give absolute
pointer positioning. The mouse port carries a quadrature encoder with no
absolute position, and the guest turns counts into pixels through its own
history-dependent acceleration curve -- which is why driving the OS 3.9
Installer headlessly stalled on 6 px radio circles. Since the Amiga
pointer *is* sprite 0, `src/pointer.rs` injects a delta, watches where
the hardware drew that sprite on the next frame, and corrects, learning
the pixels-per-count ratio rather than modelling it. Coordinates are
presented pixels, the same ones a screenshot is measured in.

Two things fell out of building it. The pointer moves in lo-res pixels,
which are two columns of the presented canvas, so not every coordinate is
exactly reachable -- hence a 2 px default tolerance and a stall detector,
so a target the guest's lattice straddles is reported as "it settles
here" instead of oscillating until the frame budget runs out. And failing
to land is an error rather than a quiet near-miss, because a script that
carried on would click the wrong thing.

## The WinUAE-inspired debugger set

**Custom-register access validator** (`[debug] validate_chipset`,
`chipset.validate`, `chipset.report`). Chipset misuse is silent on real
hardware: a write to a register the fitted Agnus does not have is decoded
and dropped, undefined bits are ignored, a byte write to a word register
latches into both halves, and a DMA pointer past Agnus's reach quietly
wraps. Every custom-register access already funnels through one place,
so all of that can be noticed and attributed to a PC or Copper address.
Findings are deduplicated by (kind, register, writer) and counted, so a
Copper list repeating a mistake every frame is one entry with a tally.

The undefined-bit masks are the OCS/ECS/AGA union and cover control
registers only -- pointers, modulos, palette entries and sprite position
words are data, so a mask there would report legitimate values. The
DMA-pointer check runs on the high half, where the intent is still
visible: by the time the pointer is formed the setters have masked the
excess away, which is exactly the bug and exactly why it is invisible.

**Per-register last-writer** (`custom.writer`) rides on the same arming:
the last value and the code that wrote it, answering "what set BPLCON3,
and from where?" without a bisect.

**Per-DMA-channel and PC-qualified watchpoints.** A watch could not see a
read at all -- the mechanism compares the word's value after each
instruction, and a DMA fetch leaves it alone, so "which sprite fetched
this word" had no way to be asked. The read-side channels now latch the
access at their own fetch sites, which are the only places that know
*which* bitplane or sprite is fetching. A watch also takes an optional PC
qualifier, for a word a dozen routines all poke.

**Self-modifying-code detector** (`[debug] detect_smc`, `smc.report`).
Reports writes landing on memory that has already executed, with the
writing PC and the distance between them, calling out a patch close
enough to sit inside the 68000's prefetch. Marking at retirement rather
than at prefetch is deliberate: the prefetch queue also reads a word past
the end of the instruction stream, and marking that would report a write
to the data following a routine as if it were code. The one uncaught
case is documented.

**Blitter, floppy and keyboard misuse checks.** The cases that hang
rather than glitch: a blit started while the previous is still running or
with its DMA off, disk DMA armed against a drive with no media or its
motor still off (the class behind the Gods and Shadow of the Beast
dead-spins), and a keyboard handshake pulse too narrow for the 6500/1 to
sample, after which the keyboard simply stops sending.

**Bus-error injection** (`fault.inject`, `fault.list`, `fault.clear`).
Exercising a guest's fault handler meant hunting for an address that
happens to be undecoded on the machine under test. A faulted access never
reaches memory, so the guest sees exactly what an undecoded address gives
it.

**Memory heat map** (`memory.heatmap`, `memory.heatmap.report`). A slot
map says what owned the chip bus at a colour clock; this says where in
memory anything is happening. A 256x256 grid over a movable window of the
address space, cells coloured by what last touched them and faded over 32
frames, reusing the per-channel attribution from the watchpoint work. The
window is movable rather than fixed at 16 MiB so the RAM a 32-bit CPU
sees above the 24-bit space can be looked at too.

## Verification

`cargo test --release` green: 1894 lib + 36 bin + 2 ctl + 24 golden
probes, no goldens re-blessed (nothing here changes timing). `cargo
clippy --all-targets --all-features --locked` and `cargo fmt --check`
clean.

The heat map was exercised against a live AROS boot over CCP: six
seconds in, 640 cells of bitplane DMA, 58 of CPU writes, 4 of Copper
fetches, against the ROM's instruction-fetch band.

## Not included

The heat map has no in-window panel -- the data model, census and PNG
export are there, but not a `Panel` variant in `ui.rs`/`window.rs`. That
is a change across two very large files and the right shape for it is
worth deciding first (a tab in the Frame Analyzer window rather than a
fourth tool window, most likely).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -75,6 +75,7 @@ repeat.
 | `--click-after SECS BUTTON MS [PORT]` | Mouse button (`left`/`right`/`middle`) for MS ms (default port 1) |
 | `--joy-after SECS BUTTON MS [PORT]` | Joystick / CD32-pad control (`up`/`down`/`left`/`right`/`red`/`blue`/...) (default port 2) |
 | `--mouse-after SECS DX DY [PORT]` | Relative mouse motion (default port 1) |
+| `--mouse-to-after SECS X Y [PORT]` | Steer the pointer to screen pixel (X, Y) via sprite 0 (default port 1) |
 | `--pot-after SECS X Y [PORT]` | Analogue stick/paddle position, 0-255 per axis (default port 2) |
 | `--insert-disk-after SECS DFN PATH` | Insert a disk image into `df0`..`df3` |
 | `--insert-cd-after SECS PATH` | Swap the CD image in the machine's CD drive (CDTV/CD32/SCSI CD-ROM) |
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -152,6 +152,7 @@ model = "68000"
 # [debug]
 # log_unmapped = "DD0000-DEFFFF"   # hex START-END, end included; or "all"
 # validate_chipset = true          # report chipset misuse with the PC that caused it
+# detect_smc = true                # report writes landing on already-executed code
 
 
 # Extended ROM image for CDTV/CD32 machines: a 256 KiB image maps at
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -151,6 +151,7 @@ model = "68000"
 # at one window.
 # [debug]
 # log_unmapped = "DD0000-DEFFFF"   # hex START-END, end included; or "all"
+# validate_chipset = true          # report chipset misuse with the PC that caused it
 
 
 # Extended ROM image for CDTV/CD32 machines: a 256 KiB image maps at
--- a/docs/debugger/console.md
+++ b/docs/debugger/console.md
@@ -60,7 +60,7 @@ Stops (each toggles: repeat the command to remove):
 | Command | Effect |
 |---|---|
 | `BREAK ADDR [COND] [IGN N]`, `B` | PC breakpoint, with the Break tab's condition grammar |
-| `WATCH ADDR [CLASS] [PC=ADDR]`, `W` | Memory word watchpoint. `CLASS` narrows it to one accessor: `CPU`, `BLITTER`, `DISK`, `COPPER`, or a DMA channel (`BPL1`..`BPL8`, `SPR0`..`SPR7`, `AUD0`..`AUD3`) -- the DMA channels catch *reads* too, which a value compare cannot see. `PC=ADDR` stops only when that instruction made the access |
+| `WATCH ADDR [CLASS] [PC=ADDR]`, `W` | Memory word watchpoint. `CLASS` narrows it to one accessor: `CPU`, `BLITTER`, `DISK`, `COPPER`, or a DMA channel (`BPL1`..`BPL8`, `SPR0`..`SPR7`, `AUD0`..`AUD3`) -- the DMA channels catch *reads* too, which a value compare cannot see. `PC=ADDR` stops only when that instruction made the access, and cannot be combined with a DMA class: only the CPU has an instruction behind an access |
 | `RWATCH NAME\|OFF`, `RW` | Custom-register write watch (`RWATCH DMACON`) |
 | `BTRAP V [H]` | Beam trap (decimal position) |
 | `CBREAK ADDR` | Copper breakpoint |
--- a/docs/debugger/console.md
+++ b/docs/debugger/console.md
@@ -60,7 +60,7 @@ Stops (each toggles: repeat the command to remove):
 | Command | Effect |
 |---|---|
 | `BREAK ADDR [COND] [IGN N]`, `B` | PC breakpoint, with the Break tab's condition grammar |
-| `WATCH ADDR [CPU\|BLITTER\|DISK]`, `W` | Memory word watchpoint; the optional filter stops only on that writer |
+| `WATCH ADDR [CLASS] [PC=ADDR]`, `W` | Memory word watchpoint. `CLASS` narrows it to one accessor: `CPU`, `BLITTER`, `DISK`, `COPPER`, or a DMA channel (`BPL1`..`BPL8`, `SPR0`..`SPR7`, `AUD0`..`AUD3`) -- the DMA channels catch *reads* too, which a value compare cannot see. `PC=ADDR` stops only when that instruction made the access |
 | `RWATCH NAME\|OFF`, `RW` | Custom-register write watch (`RWATCH DMACON`) |
 | `BTRAP V [H]` | Beam trap (decimal position) |
 | `CBREAK ADDR` | Copper breakpoint |
--- a/docs/debugger/control.md
+++ b/docs/debugger/control.md
@@ -255,6 +255,26 @@ replaced wholesale; port defaults to 2), `input.analogue {port?, x, y}`
 latches; port defaults to 2). Events drive the named port's electrical
 lines whatever device is configured there.
 
+`input.mouse_to {x, y, port?, tolerance?, max_frames?}` puts the guest's
+pointer at an absolute position instead: `x` and `y` are presented pixels
+in the same coordinate space `capture.screenshot` writes out, so a
+position read off a screenshot can be clicked directly. There is no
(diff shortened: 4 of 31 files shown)


LinuxJedi:
Force-pushed: the previous head swept ~12,600 untracked working-tree files into the mouse-servo commit (a bare `git add -A`). The branch is rebuilt with the same nine commits and identical source and docs trees; only the accidental additions are gone. 31 files changed now.

LinuxJedi:
All five review findings addressed in 877afcde:

1. `HeatMap::new` now rounds the span up to a whole number of bytes per cell (and clamps), so a span like `CELLS + 1` can no longer index past the grid. This was reachable from a control client, so it was the serious one; covered by a test for that span and for `u32::MAX`.
2. The unconstructable `0xFF` channel sentinels are gone -- the any-channel filter was half-built, so `accepts` is now exact-match and the `parse` doc describes what it actually accepts.
3. `dma_arming_obstacle` returns the `DISK_*` code instead of English that was substring-matched back into one; the wording lives in `regcheck::disk_obstacle` only.
4. Scripted `--mouse-after` deltas are held while a `--mouse-to-after` servo is steering, and the servo is advanced first so a target coming due this frame takes ownership before the deltas are considered.

On (4): it has no dedicated test. Exercising it needs a machine that really draws a sprite-0 pointer; that got as far as verifying the position mapping exactly (+32 presented pixels for +16 lo-res of HSTART) but the sprite stopped being drawn after a couple of frames, which comes down to sprite-DMA re-arm behaviour across frames. That is its own investigation rather than something to settle while fixing review comments, so I left it out and noted it in the commit.

LinuxJedi:
Both new findings fixed in 136ae871.

`WatchSource::parse` bounded every channel prefix at 8, so `aud4`..`aud7` parsed into audio channels Paula does not have. The resulting filter can never match, so the watch just never fires with nothing to explain why. Each prefix now carries the channel count the hardware actually has (bpl 1-8, spr 0-7, aud 0-3).

`run_until`'s stable-frame region only checked for `w`/`h`, so `x`/`y` alone were ignored and the run silently waited on the whole frame settling instead of the region asked for -- on a live screen that is the difference between landing and never landing. Naming any region param now requires the rest of a valid rectangle, matching `capture.region_digest`. Docs updated for both.

LinuxJedi:
All four fixed in 8e9ee0c6.

**Servo counts.** `counts_for` clamped the requested counts by the pixel error, which is a unit confusion -- the gain already is the overshoot guard, since error/gain is exactly the counts that cover this error. The clamp throttled the servo whenever the guest gives less than a pixel per count. It also returned zero counts for an error smaller than the gain, burning a frame injecting nothing and letting the stall detector call a reachable target stuck. A non-zero error now always asks for at least one count, and the tolerance and stall detector resolve it if that overshoots. Tested at 0.25 px/count.

**Keyboard latch.** Drained on every handshake edge now, with only the reporting gated on the validator, so a pulse latched before arming can no longer be reported later against an unrelated PC.

**`FrameRect::digest`** checks its own bounds arithmetic rather than trusting the parser, since the type is public and this is the method that indexes the framebuffer.

**`fault.inject`** rejects a window whose end wraps: it matched nothing, so the fault silently never fired.

LinuxJedi:
Fixed in e0cef7b7. `memory.heatmap` built its window inside a `then` closure, where a parse error had nowhere to go and `unwrap_or` swallowed it -- so a bad `base` silently became 0 instead of `-32602`. For a diagnostic that is the worst kind of wrong answer: you get a window over the wrong memory and no indication of it. Parsed before the closure now, and whether or not the call is arming, with tests.

I also swept my own diff for the same shape rather than wait for another pass -- discarded `Result`s, `.ok()`, `let _ =`. That found nothing else, but it did turn up a leftover buffered-findings array with a manual index in `note_custom_write`, from an earlier borrow arrangement that no longer applies; simplified to direct calls in the same commit.

Unrelated observation while running the suite: `config::tests::floppy_path_implies_enabled_and_write_protect_defaults` flaked once locally under parallel execution and then passed 3/3 and 4/4 on re-run. It shares `temp_adf()` with its neighbours, so it looks like a pre-existing temp-path race rather than anything on this branch -- flagging it in case it ever shows up red in CI here and gets mistaken for this work.

LinuxJedi:
All three fixed in f9a75482.

**The servo swallowed debugger stops.** `step_frame` reports `Ok` even when it ended early on a breakpoint or watch, so `mouse_to` kept stepping and the stop never surfaced -- a breakpoint on the code being driven simply would not fire while the pointer was moving. It now checks after each frame and gives up with the frames spent and how far short it stopped, leaving the stop pending for the normal path. I checked the other `step_frame` sites on this branch; they already surface stops, so this was the only one.

**Heat map window edge.** `touch` collapsed a transfer whose end fell outside the window down to its first cell, under-reporting every access near the top edge. It now marks the part that is inside, while still recording an address-space wrap at its first cell only.

**The doc comment** was my insertion landing between `set_ui_reg_watches`'s comment and the function it described, leaving one function documented as the other and the other undocumented. Both restored. Worth noting I made the same mistake earlier in this branch with `UiRegHit`/`RegWrite` and caught it only because it failed to compile -- this one was invisible to the compiler, which is exactly why it survived to review.

No test for the servo/breakpoint interaction: reaching that code needs a machine that draws a sprite-0 pointer, which is the same fixture I could not get stable earlier. Flagging it rather than papering over it.

LinuxJedi:
Both fixed in f20990bf -- they are one rule stated in two places, so it now lives in one: `WatchSource::takes_pc_qualifier`, next to the enum whose variants it is a statement about.

Combining a DMA class with `pc` installed a watch that could never fire. Only the CPU has an instruction behind an access; a DMA engine's fetch or write is issued by the chip bus with no PC to compare, so the pair describes something that cannot happen. `break.add` returns `-32602` and the console prints the reason. Either qualifier alone, and the useful `cpu` + `pc` pairing, are unaffected. Tested on both surfaces and documented on both.

For what it is worth, this is the eleventh finding of the same family on this branch -- an input accepted and then quietly not doing what it says. That is a consistent enough pattern in my own work here that I would treat the branch as wanting a careful human read before merge rather than relying on the green CI, which has now agreed with me across seventeen defects.

PR #312, opened 2026-07-28
filesys: fire MMIO doorbells on 68000's split word writes

sidick:
## Summary

Fixes a boot hang, reported via AmiAuth's on-target test suite, where any
`[[filesys]]` host-directory mount stalls the boot forever on a
68000/68010 machine (e.g. the A500 profile) — regressed between 0.11.0
and 0.12.0/0.13.0.

- `FilesysBoard::write()` only fired the `DIAG_DOORBELL`, `REG_DOSPKT`,
  and `REG_MSGPORT` MMIO doorbells on a single 32-bit access
  (`size == 4`). That holds on a 68020+'s 32-bit data bus, but a
  68000/68010 has a 16-bit external bus, so the CPU core's
  `write_move_dest_68000` splits every `move.l` destination write into
  two word bus cycles (high word, then low word) — real hardware
  behavior the core correctly reproduces.
- Because the doorbell guard never saw `size == 4` on a 68000/68010, the
  guest ROM's `move.l a0,DIAG_DOORBELL(a0)` in the DiagPoint stub never
  reached `FilesysBoard::diag_entry`, expansion init never ran, and the
  guest hung forever waiting on a host filesystem that never mounted.
  This is a regression from 497845c, which switched the services board
  from an A-line HLE trap (size-independent) to plain MMIO registers.
- Verified by reproducing the exact reported config (A500/68000, bundled
  AROS, `[[filesys]]` host-directory mount) headless: before the fix,
  boot hangs silently after `romtaginit done`, matching the bug report's
  serial capture exactly; after the fix, `HOSTFS0: handler started`
  fires, the host directory mounts, `Startup-Sequence` runs, and the
  guest reaches a shell.

## Fix

Fire each doorbell on whichever write actually completes its longword —
a single 4-byte access, or the low word of a split word pair — reading
the value back out of the window image (already latched by both halves)
rather than trusting only the write that triggered the check.

## Test plan

- [x] `cargo test` — full suite passes, including a new regression test
      (`doorbell_rings_a_packet_split_across_word_writes_like_a_68000`)
      that drives the doorbell/register sequence as two word writes, the
      way a 68000/68010 guest actually produces it.
- [x] `cargo clippy` / `cargo fmt --check` — clean.
- [x] Headless repro: A500/68000 + bundled AROS + `[[filesys]]` host
      directory mount now boots to a shell and runs `Startup-Sequence`
      instead of hanging after `romtaginit done`.
- [x] `docs/internals/peripherals.md` updated to describe the
      split-word-write handling.

### DIFF
--- a/docs/internals/peripherals.md
+++ b/docs/internals/peripherals.md
@@ -210,6 +210,15 @@ DiagPoint culls the ROM's `scsi.device` resident tag (`romtags.rs`),
 which is why setting the flag instantiates the services board even with
 no `[[filesys]]` mounts configured.
 
+These longword registers are written with a single `move.l` in the guest
+ROM/handler, but on a 68000/68010 that compiles to two word-sized bus
+cycles (high word, then low word -- a real 16-bit-bus artifact the CPU
+core reproduces). The board fires each doorbell (`DIAG_DOORBELL`,
+`REG_DOSPKT`, `REG_MSGPORT`) on whichever write actually completes the
+value -- a single 4-byte access on a 32-bit bus, or the low word of a
+split pair on a 16-bit one -- reading the result back out of the already-
+latched window image rather than trusting the write that triggered it.
+
 Amiga attributes a host filesystem cannot hold live in UAE-style `.uaem`
 sidecar files (read when present, written back on change, hidden from
 guest listings); the delete-protection bit is honoured on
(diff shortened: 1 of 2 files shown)


LinuxJedi (approved):
Great catch! Many thanks!

PR #325, opened 2026-07-28
video: redesign the reboot button glyph

warpdesign:
## Summary
- The status-bar reboot and power buttons were both green rings of the same
  size, so the two adjacent controls were easy to confuse.
- Redraws the reboot control as a golden broken-ring restart symbol with a
  bold counter-clockwise arrowhead — distinct from the green power ring in
  both shape and colour.
- Removes the now-unused `draw_cubic` bézier helper (only the old reset glyph
  used it).

## Test plan
- [x] `cargo build --release`
- [x] `cargo clippy` — no new warnings (6 pre-existing warnings unrelated to this change)
- [x] `cargo fmt --check` — clean
- [x] `cargo test --release --lib video::window` — all window/status-bar tests pass
- [x] Visual check: launch the app and confirm the reboot button shows the
      golden restart glyph, clearly distinct from the green power button beside it
- [x] Confirm clicking reboot still performs a warm reset

Before:
Screenshot 2026-07-28 at 17 21 46

After:
Screenshot 2026-07-28 at 17 22 21

### DIFF
--- a/src/video/window.rs
+++ b/src/video/window.rs
@@ -319,6 +319,7 @@ const BUTTON_GLYPH: u32 = rgba(0, 174, 0);
 const BUTTON_GLYPH_DISABLED: u32 = rgba(96, 94, 86);
 const POWER_GLYPH_ON: u32 = rgba(0, 174, 0);
 const POWER_GLYPH_OFF: u32 = rgba(150, 36, 30);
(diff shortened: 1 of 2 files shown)


LinuxJedi (approved):
Much better! Many thanks!

PR #327, opened 2026-07-28
Feature/floppybridge

hobbo91:
# Physical floppy drives: read and write real Amiga disks

Any of Copperline's four floppy bays can drive a physical 3.5" drive instead of a disk image, reading and writing real Amiga floppies over a DrawBridge, a Greaseweazle, or a Supercard Pro. Rob Smith's [FloppyBridge](https://amiga.robsmithdev.co.uk/winuae) talks to the hardware, and is vendored and compiled in, so `cargo build --release` produces a binary that drives a real drive.

The emulated machine is unchanged. The bridge only supplies the MFM the head would be passing over, so Paula, the disk DMA and `trackdisk.device` behave exactly as they do with an ADF.

## Using it

Each drive on the launcher's Floppy tab has a **Physical drive** tick box. Tick it and the media row names the interface — or `None` with nothing plugged in — with a **Configure** button leading to that drive's page for serial port, drive select, density, read mode, smart speed and read ahead. Options the chosen interface does not implement are greyed, as the driver itself reports them.

Every setting is available three ways — launcher, config file, and command line:

```toml
[floppy.df0]
bridge = "greaseweazle"      # drawbridge/greaseweazle/supercardpro/off
write_protected = true       # emulator-level protection; default true
# bridge_mode = "normal"         # normal/compatible/stalling
# bridge_port = "/dev/ttyACM0"   # omit to auto-detect
# bridge_cable = "a"             # a/b (IBM PC) or 0..3 (Shugart)
# bridge_density = "auto"        # auto/dd/hd
```

```sh
copperline --model A500 --floppy-bridge df0 greaseweazle kickstart.rom
```

plus `--floppy-bridge-port`, `--floppy-bridge-cable`, `--floppy-bridge-mode`, `--floppy-bridge-density`, `--floppy-bridge-smart-speed`, `--floppy-bridge-read-ahead` and `--floppy-bridge-writable`. A bay takes either a bridge or an image, never both.

`normal` is the default; `compatible` captures from the index, for a disk that reads badly without it; `stalling` blocks the emulated machine until each track is ready.

## Write protection

Double protection, both have to be open before anything reaches the platter: the disk's own write-protect tab, and `write_protected` in the config, which defaults to `true` as it does for an image. Both are enforced where the write would happen rather than merely reported through the drive's /WPRO line, and both come from the same reading the line is built from, so a disk the machine calls protected can never be the one being written.

## What behaves differently

- **No eject or swap** — the disk is in a real drive, changed by hand. A disk going in or coming out raises the change line.
- **No synthesized drive sounds** for that bay; the real drive makes its own.
- **Wall-clock pacing** — the platter cannot be hurried, so even a headless run is paced like an Amiga.
- **Not reproducible** — save states cannot capture the medium and input recordings will not line up. The core is as deterministic as ever; the disk under it is not.

## Vendor packaging 

Upstream ships FloppyBridge as a shared library to load at run time. Copperline links it from `vendor/floppybridge` instead. `vendor/floppybridge/README.md` records the commit vendored (`710fa15c`, reporting itself as 1.6), the licences (MPL-2.0 or GPL-2.0-or-later, both compatible with GPL-3), and every local change. `build.rs` compiles twelve sources with `cc`.

Turning the `floppybridge` feature off removes all of it rather than leaving it inert: no tick box, no `--floppy-bridge` flags in `--help` or the parser, and a config file's `bridge` keys are read and ignored so the same file stays valid across builds.

## Local changes to the vendored sources

Both are build or platform differences, not behaviour. Each is marked `COPPERLINE:` in the source and listed in the vendor README.

**Windows.** Upstream builds from a Visual Studio project that sets `NOMINMAX` and `CharacterSet: Unicode`; `cc` sets neither. Both now come from `build.rs`, and the calls that only exist under `UNICODE` are named explicitly — seven `L""` literals reaching TCHAR-generic calls, plus `SetupDiGetDeviceProperty`, which has no ANSI variant at all. The WinUAE config dialogs and a DnsQuery update check are excluded behind `FLOPPYBRIDGE_NO_GUI`, so the build needs no resource compiler. Amiberry cuts the same seam.

**Linux — likely an upstream bug.** Opening a tty blocks until carrier detect is asserted unless `CLOCAL` is already set, and the sources set `CLOCAL` immediately *after* opening. A USB CDC interface that never raises DCD — a Greaseweazle among them — hangs in `open()` forever. Upstream sidesteps this on macOS by passing `O_NDELAY` and nowhere else, which is why it only appears on Linux. Confirmed by `stty -F /dev/ttyACM0 clocal`, which turns the hang into a normal handshake. Worth reporting upstream; it appears to be latent in Amiberry's vendored copy too but would need to test. 

## Testing

Verified against real hardware (the best that I can) — a Greaseweazle V4.1 and a Workbench 1.3 disk — on macOS throughout, and on Windows 11 on ARM. Reads were A/B tested against the same disk as an ADF. A hardware probe test decodes the captured MFM and asserts eleven AmigaDOS sector headers in order, so the stream crossing the FFI is provably the disk's. A second probe reports the drive's status lines with the motor stopped and running:

```sh
cargo test --release floppybridge_status_probe -- --ignored --nocapture
```

Both feature configurations build and test clean, with clippy and rustfmt clean in each.

## Still to be tested

**Linux.** On the only Linux machine available — Ubuntu 24.04 in a Parallels VM on Apple silicon — the Greaseweazle works for one session and then refuses to talk until physically replugged. This is not Copperline's doing: the vendor's own `gw info` reproduces it identically with Copperline never involved, no software reset recovers it, and `dmesg` fills with `xhci_hcd ... Event dma ... not part of TD`. The same board is fine natively on macOS and under the same hypervisor on Windows. It wants confirming on real Linux hardware before Linux is called done.

**Other interfaces.** Only a Greaseweazle has been on the end of this. DrawBridge and Supercard Pro are wired up and their driver-reported capabilities drive the UI, but neither has been exercised against real hardware.

Screenshot 2026-07-29 at 00 33 34
Screenshot 2026-07-29 at 00 33 45
Screenshot 2026-07-29 at 00 33 57

### DIFF
--- a/src/floppybridge/mod.rs
+++ b/src/floppybridge/mod.rs
@@ -1,0 +1,810 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+//! Real floppy drives, through Rob Smith's FloppyBridge library.
+//!
+//! A *bridge* replaces one Amiga drive's disk image with a physical 3.5"
+//! drive attached over a DrawBridge, a Greaseweazle, or a Supercard Pro. The
+//! emulated machine is unchanged: the bridge only supplies the MFM the head
+//! would be passing over, so Paula, the disk DMA, and trackdisk.device all
+//! behave exactly as they do with an image.
+//!
+//! # How it attaches
+//!
+//! The library reads the disk continuously on its own thread and keeps the
+//! track under the head captured and ready. [`Bridge::read_track`] takes a
+//! whole finished revolution from it in one call, packed MFM plus the length
+//! it wrapped at -- which is the shape a captured revolution already has in
+//! [`crate::floppy`], so the existing rotation, PLL, and sync-word machinery
+//! reads a real disk with no special case in the hot path.
+//!
+//! Nothing here waits on the drive. A track that has not been captured yet
+//! comes back as `None` and the caller asks again, exactly as it already did
+//! while the motor spun up. Blocking instead would stop the emulated Amiga --
+//! CPU, sprites, pointer and all -- every time the head moved, which is the
+//! one thing a real drive never does to a real Amiga.
+//!
+//! Because a revolution is served whole, it has to start where the real one
+//! does. That is why the bridge is opened in the library's `Compatible` mode,
+//! which captures from the index: a stream anchored anywhere else would wrap
+//! in the middle of a sector rather than in the gap between two, and the guest
+//! would see disk errors. Upstream's faster index-less modes suit an emulator
+//! that streams cells continuously off the drive, which this is not.
+//!
+//! Writing goes the same way round: [`Bridge::write_track`] hands the MFM over
+//! a word at a time at the rotational position the head would be passing, as a
+//! real drive lays cells down, and commits it to the platter -- without waiting
+//! for the platter to take it, for the same reason reads do not wait.
+//!
+//! # Where it comes from
+//!
+//! FloppyBridge is compiled into the emulator from `vendor/floppybridge`, so a
+//! build that says it supports a physical drive can actually drive one, with
+//! nothing to download and nothing to install. Upstream ships it as a shared
+//! library to be loaded at run time; Copperline links it instead, because a
+//! user should not have to fetch a second file to use a feature the build
+//! claims to have.
+//!
+//! Keeping upstream current is a maintainer's job and close to a wholesale
+//! copy: `vendor/floppybridge/README.md` records the commit vendored and the
+//! handful of local changes, all of them Windows or Linux build differences
+//! rather than behaviour.
+//!
+//! Turning the `floppybridge` Cargo feature off, which is on by default,
+//! compiles all of this out -- the C++ included.
+//!
+//! # Determinism
+//!
+//! A real drive spins in wall-clock time and the disk under it can be swapped
+//! by hand, so a run using a bridge is *not* reproducible: save states cannot
+//! capture the medium, and a replayed input recording will not line up. Nor can
+//! the machine be run faster than the platter turns, so a bridged machine is
+//! paced to wall-clock time even in a headless run that would otherwise be
+//! unthrottled -- see `main`. As with NAT networking, the constraint is
+//! documented rather than enforced by a flag: the emulated core is as
+//! deterministic as ever, it is the disk under it that is not.
+
+mod ffi;
+
+use std::ffi::{c_char, c_int, c_uint, c_void, CStr, CString};
+use std::sync::{Mutex, OnceLock};
+
+use log::warn;
+
+pub use ffi::{config_option, BridgeDensityMode, BridgeMode, DriveSelection};
+
+/// The largest string the library will write into a caller-provided buffer.
+/// Upstream's own `BRIDGE_STRING_MAX_LENGTH`.
+const STRING_MAX: usize = 255;
+
+/// Bytes to allow for one captured revolution. Upstream sizes its own track
+/// buffers at `MFM_BUFFER_MAX_TRACK_LENGTH` (0x3A00 * 2), so nothing longer
+/// than this can come back: a DD revolution is around 12.7K of MFM and an HD
+/// one twice that, leaving plenty of room for a drive running off-speed.
+const MAX_TRACK_BYTES: usize = 0x3A00 * 2;
+
+/// Serialises calls into the library itself.
+///
+/// The `BRIDGE_*` half works through state the library owns process-wide: the
+/// port scan fills a file-scope vector on the sizing call and reads it back on
+/// the next, and the driver query hands out pointers into storage that belongs
+/// to the library.
+/// None of that survives two threads at once -- reliably enough that the test
+/// suite aborted with SIGABRT whenever two of these ran together -- and the
+/// launcher does exactly that, asking what is plugged in while a machine is
+/// running.
+///
+/// Per-drive `DRIVER_*` calls are deliberately not covered: each belongs to one
+/// [`Bridge`] owned by the emulation thread, and the library serialises its own
+/// worker behind them. Only [`Bridge::open`] takes this, because creating a
+/// driver goes through the `BRIDGE_*` half.
+static LIB_LOCK: Mutex<()> = Mutex::new(());
+
+/// Take the library lock, ignoring poisoning: a panic in another thread's
+/// query leaves the library no more broken than it already was, and refusing
+/// to look at the hardware ever again would be the worse failure.
+fn lib_lock() -> std::sync::MutexGuard<'static, ()> {
+    LIB_LOCK.lock().unwrap_or_else(|e| e.into_inner())
+}
+
+/// Copy a bridge-owned C string. The pointers handed back belong to the bridge
+/// and several are documented as invalid after the next call, so everything
+/// crossing into Copperline is copied immediately.
+///
+/// # Safety
+/// `ptr` must be null or a valid NUL-terminated string.
+unsafe fn owned_string(ptr: *const c_char) -> String {
+    if ptr.is_null() {
+        return String::new();
+    }
+    CStr::from_ptr(ptr).to_string_lossy().into_owned()
+}
+
+/// The bridge's version, as `(major, minor)`.
+///
+/// Asked once and remembered: what is linked in cannot change under us, and
+/// the launcher puts this in a heading.
+pub fn version() -> Option<(u32, u32)> {
+    static VERSION: OnceLock<Option<(u32, u32)>> = OnceLock::new();
+    *VERSION.get_or_init(|| {
+        let _guard = lib_lock();
+        let mut out: *mut ffi::BridgeAboutRaw = std::ptr::null_mut();
+        // `false`: never let the bridge phone home for an update check.
+        unsafe { ffi::BRIDGE_About(false, &mut out) };
+        if out.is_null() {
+            return None;
+        }
+        let about = unsafe { &*out };
+        Some((about.major_version as u32, about.minor_version as u32))
+    })
+}
+
+/// The physical drive an interface reports being attached to.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum DriveType {
+    /// A plain 3.5" double-density drive: the normal Amiga case.
+    Dd35,
+    /// A 3.5" high-density drive, which can also read DD media.
+    Dd35Hd,
+    /// A 5.25" single-density drive.
+    Sd525,
+}
+
+/// One driver the library offers, by index.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DriverInfo {
+    pub index: u32,
+    pub name: String,
+    pub manufacturer: String,
+    pub url: String,
+    /// Bitmask of [`config_option`] flags this driver honours.
+    pub config_options: u32,
+}
+
+impl DriverInfo {
+    pub fn supports(&self, option: u32) -> bool {
+        self.config_options & option != 0
+    }
+}
+
+/// Every driver the loaded library offers (DrawBridge, Greaseweazle, ...).
+/// Empty when no library is loaded.
+pub fn drivers() -> Vec<DriverInfo> {
+    let _guard = lib_lock();
+    let count = unsafe { ffi::BRIDGE_NumDrivers() };
+    (0..count)
+        .filter_map(|index| {
+            let mut raw: *mut ffi::BridgeDriverRaw = std::ptr::null_mut();
+            if !unsafe { ffi::BRIDGE_GetDriverInfo(index, &mut raw) } || raw.is_null() {
+                return None;
+            }
+            let d = unsafe { &*raw };
+            Some(DriverInfo {
+                index,
+                name: unsafe { owned_string(d.name) },
+                manufacturer: unsafe { owned_string(d.manufacturer) },
+                url: unsafe { owned_string(d.url) },
+                config_options: d.config_options as u32,
+            })
+        })
+        .collect()
+}
+
+/// Whether an interface the library recognises is plugged in right now.
+///
+/// The port scan below only reports ports the library believes carry a
+/// supported board, so an empty list means there is nothing to drive a real
+/// disk with. Scanning walks the host's serial bus, so this is sampled at
+/// deliberate moments -- a bay being switched over to a real drive -- rather
+/// than every frame.
+pub fn interface_connected() -> bool {
+    let _guard = lib_lock();
+    !com_ports_locked().is_empty()
+}
+
+/// Serial ports the library can see, for the drivers that need one named.
+pub fn com_ports() -> Vec<String> {
+    let _guard = lib_lock();
+    com_ports_locked()
+}
+
+/// The scan itself. Split out because the caller may already hold the lock,
+/// which is not re-entrant.
+fn com_ports_locked() -> Vec<String> {
+    // Upstream fills a caller-provided buffer with NUL-separated names and
+    // reports the size it wanted; ask twice so a long list is never truncated.
+    let mut size: c_uint = 0;
+    unsafe { ffi::BRIDGE_EnumComports(std::ptr::null_mut(), &mut size) };
+    let cap = (size as usize).max(STRING_MAX * 8);
+    let mut buffer = vec![0u8; cap];
+    let mut size = cap as c_uint;
+    if !unsafe { ffi::BRIDGE_EnumComports(buffer.as_mut_ptr() as *mut c_char, &mut size) } {
+        return Vec::new();
+    }
+    buffer
+        .split(|&b| b == 0)
+        .filter(|s| !s.is_empty())
+        .map(|s| String::from_utf8_lossy(s).into_owned())
+        .collect()
+}
+
+/// Everything needed to open one real drive.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct BridgeConfig {
+    /// Which driver to open, by the index [`drivers`] reported it at.
+    pub driver: u32,
+    pub mode: BridgeMode,
+    pub density: BridgeDensityMode,
+    pub drive: DriveSelection,
+    /// Which serial port the interface is on. `None` -- the default -- lets the
+    /// driver find its own device, which every current driver supports
+    /// ([`config_option::AUTO_DETECT_COMPORT`]); name one to pin it, which
+    /// matters when two interfaces are plugged in at once.
+    pub port: Option<String>,
+    pub smart_speed: bool,
+    pub auto_cache: bool,
+}
+
+/// One open real drive.
+///
+/// Not `Clone`: the handle owns a physical device. Dropping it closes and frees
+/// the driver, which parks the motor.
+pub struct Bridge {
+    handle: ffi::BridgeDriverHandle,
+    max_cylinder: u8,
+    /// Scratch for whole-track reads, kept between calls so polling for a
+    /// track does not churn a 29K allocation each time round.
+    capture: Vec<u8>,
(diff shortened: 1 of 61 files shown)


LinuxJedi:
Getting local Fable and codex to review this, as it is too big for Copilot.

LinuxJedi (commented):
_Review by Claude (Fable 5, via Claude Code), at the maintainer's request._

## Overview

Any of the four floppy bays can be backed by a real 3.5" drive over a DrawBridge, Greaseweazle, or Supercard Pro. FloppyBridge (~17k lines of C++) is vendored in `vendor/floppybridge` and compiled in by `build.rs`; a new `src/floppybridge/` module (raw FFI + safe wrapper) feeds captured revolutions into the existing `TrackRev` machinery in `src/floppy.rs`, so Paula/DMA/trackdisk see a real disk exactly as they see an ADF.

This is a very strong PR: the architecture is right (the bridge supplies MFM, the emulated machine is untouched -- fully in line with the hardware-first rule), write protection is genuinely defense-in-depth and enforced at the write site, the non-blocking read design is correct, and the docs are excellent. I verified the vendored tree against upstream commit `710fa15c`: 25 of 31 files are byte-identical, and the six that differ match the README's account (TCHAR->`*A` calls, the `FLOPPYBRIDGE_NO_GUI` seam, the Linux `CLOCAL` open fix -- which does look like a real upstream bug worth reporting). All 11 runnable bridge tests pass locally, and the vendored C++ builds clean on macOS (Apple silicon).

I found two functional issues worth fixing before merge, plus a set of minor items.

## Issue 1: media polling is gated on the guest acknowledging VERTB

In `src/bus.rs` (`flush_pending_vbi`), `poll_bridge_media()` sits inside `if self.paula.intreq & INT_VERTB == 0`, so it only runs when VERTB transitions from acknowledged to raised. A guest that never writes INTREQ to clear VERTB -- trackloaders and demos that disable interrupts and poll the beam, i.e. exactly the software people will point a real disk at -- leaves the bit set forever, and disk-change detection, write-protect-tab refresh, and unplug detection all stop after the first frame. A disk swapped by hand would never raise the change line.

**Fix:** move the call to the top of `flush_pending_vbi()`, outside the INTREQ guard -- it is already paced to once per frame wrap by `pending_vbi`.

## Issue 2: the "same revolution" filter in `ensure_track` is dead code

In the spent-revolution path (`src/floppy.rs`, `ensure_track`), the order is:

1. `drive.cached = CachedTrack::default()` (filler is not in place on the first spent pass, so this always fires);
2. `let previous = spent.then(|| drive.cached.revs.first().map(|r| r.words.clone()))` -- computed from the **just-cleared** cache, so it is always effectively `None`;
3. `.filter(|(words, _)| previous.flatten().is_none_or(|p| p != *words))` -- always passes.

I checked the vendored side to see whether the driver saves you: it does not. `internalSwitchCylinder` (CommonBridgeTemplate.cpp:729) leaves `current` unchanged (and still `ready`) when no new capture has landed, and `getMFMTrack` happily returns it again. So when the background capture loses the race with the guest's rotation -- roughly a coin flip, since both take about one revolution -- the guest is handed the same non-index-aligned recording again, which is precisely the mid-sector splice the comment says this filter exists to prevent. It self-heals a revolution later, so the symptom is intermittent sector errors and trackdisk retries in the default `normal` mode, not a hard failure -- but it partially defeats the feature's central design idea.

Related: `bridge.switch_buffer(side)` is called on **every** 1 ms poll while `spent` is set. `internalSwitchCylinder` swaps `current`/`last` on each call when only `last` is ready, so repeated polling can serve revolution N-1 after N.

**Fix:** capture `previous` before clearing the cache (or compare against `bridge_tracks[track]`, which still holds the spent recording), and call `switch_buffer` once per spent revolution rather than once per poll attempt.

## Minor issues

- **`--floppy-bridge df0 off` clears the config file's image.** `ConfigOverrides::apply` unconditionally does `drive.path = None; drive.paths = None` for any `--floppy-bridge` value, including `off`, whose whole point is "go back to images". `off` should skip the path clearing.
- **Stale doc comment on `RawFloppyDrive::bridge_mode`** (config.rs): says values are "`fast`, `compatible` (the default...), `turbo`, or `stalling`" -- but the default is `normal`, `turbo` is rejected by name, and the accepted spellings are `normal`/`fast`/`compatible`/`stalling`. `bridge_smart_speed` is also described differently in the typed struct ("track a disk whose data rate wanders") vs the raw struct ("switch to turbo between accesses" -- the one that matches upstream).
- **Two misplaced doc comments:** `FloppyDrive::is_bridged` carries `has_media`'s paragraph glued above its own one-liner, and in window.rs `media_bar`'s doc comment now sits on `drive_is_bridged`. Plus a doubled "Resolve the driver by name..." comment in `attach_floppy_bridges`, and `a_dropped_bridge_hands_the_device_back`'s doc starts with a leftover sentence referencing a nonexistent `Source::Driver`.
- **Stale "installed library" phrasing** from the earlier dynamic-loading design: `drivers()` says "Empty when no library is loaded" (it is linked in; the PR's own test asserts non-empty), `bridge_driver_supports` has a dead `drivers.is_empty()` branch justified by "a machine that simply has not installed it yet", and `floppybridge_inventory` is ignored with reason "needs a FloppyBridge library installed" (it needs nothing -- it could run un-ignored).
- **Docs vs code log strings:** `docs/guide/floppybridge.md` quotes `floppy.df0 real drive attached:` and "the installed FloppyBridge has no X driver"; the code logs `physical drive attached` and "the built-in FloppyBridge has no X driver". Grep-ability matters for troubleshooting docs.
- **Vendor provenance nit:** `resource.h` does not match the claimed commit `710fa15c` (it has `IDB_DONATE`/`IDB_SMALLLOGO*` where that commit has `IDB_PATREON` -- it looks like it came from a different revision, probably the WinUAE tree). Harmless since the GUI code that includes it is compiled out, but the README's "vendored from commit `710fa15c`" claim should be accurate; the README also says local changes are "confined to four files" when six files differ.

## Quality notes (no action needed)

- The FFI layer is careful: the `BRIDGE_*` process-wide-state lock is documented, justified by an observed SIGABRT, and covered by a regression test; `read_track` clamps the reported bit count to the buffer rather than trusting the driver; `MAX_TRACK_BYTES` matches upstream's `MFM_BUFFER_MAX_TRACK_LENGTH` (checked); the update check is permanently disabled (`BRIDGE_About(false)`); `Drop` takes the lock and the failed-open path frees the half-built driver.
- Write protection is exemplary: config + tab, checked at the write site from the same readings the /WPRO line was built from, protected-by-default in config, parser, launcher, and `Default` (with a comment explaining why `derive(Default)` would be wrong). `Bridge::write_track` re-asks the drive, which technically contradicts the "no second reading" comment in floppy.rs, but that layer can only refuse, never allow -- safe direction.
- Forcing wall-clock pacing for headless runs with a bridge, save-state fields `#[serde(skip)]` (no `STATE_VERSION` bump needed -- serialized shape unchanged), power-off releasing the interface with re-attach on power-on and on failed launcher rebuilds: all thought through.
- Feature gating is thorough on both sides (`--help`, parser, launcher tick box, UI page all absent without the feature; config keys read-and-ignored so files stay portable). All new code is ASCII-clean.

## Test coverage

Config parsing, CLI overrides, launcher greying/round-trips, and the status bar are well covered, and the hardware-gated probe tests (MFM sector-header decode, status lines, device release on drop) are a genuinely good idea. The gap: the most intricate new emulator logic -- the spent-revolution/filler/retry machinery in `ensure_track`, i.e. exactly where Issue 2 lives -- has no automated coverage, because `Bridge` is a concrete FFI type with no trait seam to fake. Worth considering a small trait (or test-only injection of captured revolutions) so that logic can be exercised with a scripted "drive"; it would have caught Issue 2.

**Verdict:** looks great once Issues 1 and 2 are addressed -- both are small, localized fixes. The minor items are comment/doc cleanups that could land in the same push.

src/floppybridge/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+        let cylinder = self.clamp_cylinder(cylinder);
+        unsafe {
+            ffi::DRIVER_setSurface(self.handle, side);
+            for (i, word) in words.iter().enumerate() {
+                ffi::DRIVER_writeShortToBuffer(
+                    self.handle,
+                    side,
+                    cylinder as c_uint,
+                    *word,
+                    (start_bit + i * 16) as c_int,
```
LinuxJedi, line 522: [P1] Preserve the requested phase for partial physical writes

`mfmPosition` is not used to position the physical write. The vendored `writeShortToBuffer` records only the first position, and `commitWriteBuffer` reduces that to a `writeFromIndex` boolean (within 30 bits of index). The queued worker later calls `writeData` at the platter phase current when it handles the command. Thus any non-index/partial guest DMA can overwrite an unrelated sector or protection area instead of `start_bit`. This needs phase synchronization, or the words need to be streamed while DMA advances.

src/floppybridge/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+    }
+}
+
+/// Push the settings onto a driver that has been created but not yet opened. A free function because it runs before
+/// there is a [`Bridge`] to call it on.
+fn apply(handle: ffi::BridgeDriverHandle, config: &BridgeConfig) {
+    unsafe {
+        ffi::BRIDGE_DriverSetMode(handle, config.mode as u8);
+        ffi::BRIDGE_DriverSetDensityMode(handle, config.density as u8);
+        ffi::BRIDGE_DriverSetCable2(handle, config.drive as u8);
```
LinuxJedi: [P1] Propagate unsupported cable selections

This setter can return false. For example, SuperCard Pro advertises only `CONFIG_OPTIONS_DRIVE_AB`; selecting Shugart 0–3 is rejected by the vendored setter. Because the result is ignored, the config remains at its default Drive A and `BRIDGE_Open` can still succeed, silently reading or writing a different physical disk than requested. Validate the driver capability or turn setter failure into an open/configuration error.

src/main.rs
```diff
@@ -1605,7 +1688,17 @@ fn main() -> Result<()> {
...
         || cli.control.is_some();
-    let paced = !headless_capture;
+    // A real drive on a bridge is the exception: its platter turns in
+    // wall-clock time and cannot be hurried. Left unthrottled, the emulated
+    // machine outruns it -- spinning the motor up and down faster than it can
+    // reach speed, and stepping past tracks before the drive has captured
+    // them -- so the guest sees a drive that answers almost nothing. Pace a
+    // bridged machine like a real Amiga, whatever else was asked for.
+    let bridged = cfg.floppy.bridges.iter().any(Option::is_some);
+    let paced = !headless_capture || bridged;
```
LinuxJedi, line 1698: [P2] Keep bridged sessions paced in every runner

This initial `paced` choice is immediately overridden in `run_headless_benchmark`, `gdbstub::run`, and `control::headless::run`, each of which calls `set_paced(false)`. GUI `toggle_warp` can do the same. Those paths let the emulated machine outrun the physical platter again. Preserve the bridged invariant inside every runner and disable warp for bridged bays, or reject those incompatible modes explicitly.

src/video/window.rs
```diff
@@ -4379,6 +4396,12 @@ impl App {
                 self.ui.menu_scroll = 0;
                 self.request_redraw();
             }
+            // A bridged drive's media is a real disk in a real drive: it is
+            // loaded, swapped, and ejected by hand. The buttons stay drawn so
+            // the drive is visibly there and numbered, but they do nothing.
+            BarControl::DriveLoad(idx) if self.drive_is_bridged(idx) => {}
+            BarControl::DriveSwap(idx) if self.drive_is_bridged(idx) => {}
+            BarControl::DriveEject(idx) if self.drive_is_bridged(idx) => {}
```
LinuxJedi, line 4404: [P2] Enforce bridge/image exclusivity at the controller API

These guards cover only status-bar buttons. Scheduled `--insert-disk-after`, drag-and-drop, and control-protocol insertion still call `FloppyController::insert_disk_image` for a bridged bay. `insert_image` leaves `drive.bridge` attached, so bridge reads keep winning while the callers report the image mount as successful. Reject image insertion and ejection when `drive.bridge.is_some()` (or explicitly detach it) in the controller so every route observes one invariant.

LinuxJedi (commented):
_Review by OpenAI Codex, submitted through the maintainer’s GitHub account at their request._

I deduplicated this against Claude’s review and have not repeated its findings about VERTB-gated media polling, stale capture retirement, or the `--floppy-bridge … off` image-path override.

This review adds four functional issues: two P1 hardware-safety/correctness problems and two P2 integration problems. The inline comments contain the details. These should be addressed before merge.

src/floppybridge/mod.rs
hobbo91: Confirmed, and this was the most serious finding — it can corrupt a real disk. Traced it right through: `commitWriteBuffer` reduces the per-word position to a `writeFromIndex` boolean (the 30-bit window), and `writeCurrentTrackPrecomp` turns that into `cue_at_index`. There is no third option, so `start_bit` genuinely goes nowhere.

Fixed by refusing what cannot be placed rather than gambling with it: a write is allowed if it starts at the index (upstream's own 30-bit slack) or if it covers a whole revolution, which lands correctly wherever it begins. Anything else is refused with a log line naming the position and the track length. AmigaDOS writes all eleven sectors at once, so this should not fire in ordinary use — and writes are confirmed still working on real media after the change.

Full phase-accurate partial writes would need the words streamed as DMA advances, which is a bigger change than I want to make without hardware verification. Refusing is the safe direction meanwhile.

src/floppybridge/mod.rs
hobbo91: Confirmed and fixed. `apply()` now reports which settings the driver refused, and a refused drive select fails the open with a named error instead of silently leaving the driver on Drive A. The other settings warn and fall back to the driver default, which is the right severity for preferences — the drive select is not one, since it decides which physical drive is being spoken to.

Your Supercard Pro example is exactly the case: it advertises only `CONFIG_OPTIONS_DRIVE_AB`, so a Shugart selection was being dropped on the floor.

src/main.rs
hobbo91: Confirmed and fixed, one level lower than you suggested: `set_paced` itself now refuses to unpace a machine with a bridged bay. That covers all four `set_paced(false)` sites — benchmark, GDB stub, control server, headless — plus GUI warp, in one place, so a future runner cannot quietly opt out of the invariant.

src/video/window.rs
hobbo91: Confirmed and fixed. `insert_disk_image`, `insert_disk_image_bytes` and `eject_disk_image` all reject a bridged bay now, so scheduled `--insert-disk-after`, drag-and-drop and the control protocol observe the same invariant the status bar does, instead of reporting a successful mount that the bridge then overrides.

hobbo91:
Thanks both — these were accurate and useful reviews. I checked every claim against the code and all but one held up. Everything below is pushed as of `1a5ee0c`.

## Issue 1 — media polling gated on VERTB

Confirmed and fixed. `poll_bridge_media()` now runs before the INTREQ test, still paced once per frame by `pending_vbi`. You are right about who this bites: software that disables interrupts and polls the beam is exactly what a real disk gets pointed at, and it left disk-change detection, write-protect refresh and unplug detection dead after the first frame.

## Issue 2 — the "same revolution" filter is dead code

Confirmed: `drive.cached` is emptied before `previous` is taken from it, so the filter never fires.

Making it live is worse, and I have hardware evidence. With the filter working, Workbench 1.3 does not boot at all — the drive cannot finish a capture in every revolution the guest spends reading the last one, so refusing the stale recording hands the guest filler and starves it. The log shows it stuck retrying track 0. Serving the stale recording costs one spliced sector and a retry; ten good sectors out of eleven beat none.

Your related point about `switch_buffer` per poll is right too. With no `next` ready, `internalSwitchCylinder` swaps `current`/`last` and returns N-1. I tried throttling retirement to one revolution: about 2s per capture, still no boot. There is no way to ask "is a new revolution ready?" without rotating the buffers, so N+1 and N-1 cannot be told apart.

So the filter is removed rather than left claiming a guarantee it never had, and the code and the guide now state the trade instead: `normal` is faster and can splice when the drive falls behind, `compatible` is the mode with no seam. Measured after: ~14s of drive wait and a reliable desktop, 3/3 runs.

## Minor items

- **`--floppy-bridge dfN off` clearing the image path** — fixed; `off` now skips the clearing, since handing the bay back to images is the one thing it exists to do.
- **Stale `bridge_mode` doc** — fixed. Now `normal` (default), `compatible`, `stalling`, with `fast` accepted as the driver's own spelling for `normal`, and `turbo` refused by name.
- **`bridge_smart_speed` described two ways** — you were right that the raw struct was the accurate one. Reading `checkSmartSpeed` in the vendored source: it times each track and marks the ones with a *uniform* rate as safe to offer faster, so the typed struct's "data rate wanders" had it backwards. Both now say that, and note that we derive cell timing from revolution length so the per-cell speed goes unused.
- **Misplaced doc comments** — all four fixed: `has_media`'s paragraph returned to `has_media`, `media_bar`'s to `media_bar`, the doubled "Resolve the driver by name" removed, and the leftover `Source::Driver` sentence deleted.
- **Stale "installed library" phrasing** — fixed in all three places.
- **Docs vs code log strings** — fixed; the guide now quotes `physical drive attached` and "the built-in FloppyBridge has no *X* driver". Agreed that grep-ability is the point of those.

## Vendor provenance — half right

Thanks for checking this. **`resource.h` does match `710fa15c`.** I cloned upstream at that commit and diffed: it has `IDB_DONATE`/`IDB_SMALLLOGO`, which is exactly what is vendored. `IDB_PATREON` must come from a different revision.

The file count was wrong in both directions, though: **five** files differ, not four as the README claimed nor six. They are `ArduinoFloppyBridge.cpp`, `FloppyBridge.cpp`, `GreaseWeazleBridge.cpp`, `SerialIO.cpp` and `ftdi.cpp`. README corrected, and it now records that every other vendored file is byte-identical to that commit.

## Test coverage

Agreed, and Issue 2 makes the case better than I could — that logic had no automated coverage and the bug lived in it. I have not added the trait seam in this PR: the behaviour I would be pinning down has just changed, and a fake encoding the wrong model would be worse than none. Worth a follow-up issue.

## Unrelated, for the record

While in here I also chased the drive noise and slowness on a real drive, and found no defect on our side: the bridged path issues 1864 guest stepper pulses over 1863 cylinders on a Workbench 1.3 boot, byte-identical to the same boot from an ADF. Head travel, motor cycling and step pattern all match the image path, which is the emulator's model of a real Amiga. I tried moving the head only when a track is read — 93% fewer seek commands for the same travel — and it made no audible difference and lost the idle no-disk click, so it is reverted and the tree is unchanged by it.

Amiberry was used as the comparison for most of this work, and that can load up WB1.3 on a stock Amiga around 15 seconds faster than Copperline can, where as an ADF of the same image/config is basically identical between both emulators - put *A LOT* of time into tracking down why but ultimately came up short of a definitive answer. The next step for this would be to get a full debug log in Amiberry to see where the extra time is going. 

**Still untested: DrawBridge and Supercard Pro hardware, and Linux on a non-virtualised host.**

LinuxJedi (approved):
Fantastic work! Many thanks for this. There may well be some minor bugs left, but I'll pull this in now and they can be fixed as they are found.

hobbo91:
> Fantastic work! Many thanks for this. There may well be some minor bugs left, but I'll pull this in now and they can be fixed as they are found.

Thanks! and yes 100% expecting there to be bugs / issues I've not factored in once this gets a bit more testing, especially as I am only limited to a greaseweazle.

Issue #267, opened 2026-07-23
[Bug]: Crash when running AROS ROM


warpdesign:
[copperline-crash-aros-roms.zip](https://github.com/user-attachments/files/30322283/copperline-crash-aros-roms.zip)

### Area

Crash, hang, or freeze

### Summary

Copperline crashes

### Copperline version

0.12.0

### Host system

macOS 15

### Machine/configuration

```toml
macOS 15 Macbook Air M4
```

### Reproduction steps

Start copperline with the following parameters:
> /Applications/Copperline.app/Contents/MacOS/copperline --config /tmp/aros-no-media.toml --noaudio --screenshot-after 10 /tmp/copperline-host-crash-repro-2.png

### Actual behaviour

Copperline crashes

### Expected behaviour

Copperline should not crash

### Logs or debugger output

```text
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               copperline [95060]
Path:                  /Applications/Copperline.app/Contents/MacOS/copperline
Identifier:            dev.copperline.Copperline
Version:               0.12.0 (0.12.0)
Code Type:             ARM-64 (Native)
Parent Process:        codex [27213]
Responsible:           ghostty [7686]
User ID:               501

Date/Time:             2026-07-23 22:10:01.8161 +0200
OS Version:            macOS 15.5 (24F74)
Report Version:        12
Anonymous UUID:        EBDA6BE7-B2D9-157A-A9BA-F5033F1CA693

Sleep/Wake UUID:       031F2F58-3E66-4E24-8D77-210977AB407B

Time Awake Since Boot: 350000 seconds
Time Since Wake:       323 seconds

System Integrity Protection: enabled

Crashed Thread:        0  main  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   copperline [95060]

Application Specific Information:
abort() called


Thread 0 Crashed:: main Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x189379388 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1893b288c pthread_kill + 296
2   libsystem_c.dylib             	       0x1892bbc60 abort + 124
3   HIServices                    	       0x1902039e4 _RegisterApplication + 12732
[...]

PR #331, opened 2026-07-29
ci: functional testing on all three platforms plus artifact boot smokes

LinuxJedi:
## What

Windows and macOS packaging jobs only proved the code compiles; no shipped artifact was ever executed, and Linux had one thin smoke whose premise had gone stale. This PR adds functional testing on all three platforms and boots every artifact we publish.

- **windows.yml**: after building the zip, expands it outside the checkout and boots it from a clean directory (validates the sibling `aros\` ROM lookup and the statically linked CRT on a bare host), then runs `cargo test --release --locked --target x86_64-pc-windows-msvc` — the inline unit suites plus the probe golden renders, reusing the zip step's dependency build. Probe diff masks upload on failure and the release-attach step is now gated behind all of it. Checkout forces LF since the repo has no `.gitattributes`.
- **macos.yml**: mounts the freshly built dmg and boots the .app straight off the read-only volume from an empty directory. Asserts both `lipo` slices and the ad-hoc signature, runs arm64 natively and x86_64 under Rosetta — the first time the Intel half of the universal binary is executed anywhere.
- **ci.yml**: new `linux-tests` job on ubuntu running the unit suites and the probe goldens. No display or GPU stack needed: capture runs use the windowless path, so only the ALSA/udev build deps are installed.
- **appimage.yml**: boots the finished AppImage from an empty directory, proving the `usr/share/copperline/aros` lookup.
- **linux-present.yml**: the smoke stopped testing presentation when the windowless capture path landed (#267) — a `--screenshot-after` run never touches winit/wgpu. It now starts a real windowed session under xvfb via `--control-gui` and drives `capture.screenshot` + `shutdown` over CCP, so lavapipe Vulkan surface creation, live frame presentation, and a clean event-loop exit are genuinely exercised again.

## Why the smokes run from an out-of-tree cwd

`romsearch.rs` falls back to cwd-relative `assets/aros`, so a repo-root cwd would mask a broken bundled-ROM lookup. Verified locally: the staged layout boots from an empty cwd, and removing the bundled ROM makes the run exit 1, so the smoke genuinely catches packaging regressions. The CCP screenshot/shutdown sequence was also rehearsed end-to-end locally (clean exit 0).

## Note for the first run

The probe goldens were blessed on macOS. If the Linux or Windows jobs show pixel diffs, that is a real cross-platform divergence in the core — exactly the signal these jobs exist to surface — and the diff masks will be in the uploaded artifacts.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -66,6 +66,9 @@ jobs:
         # (libfuse2 for the AppImage runtime is installed above).
         run: |
           appimage="$PWD/${{ steps.artifact.outputs.path }}"
+          # linuxdeploy marks the image executable; do not rely on that
+          # staying true across tooling changes.
+          chmod +x "$appimage"
           smoke="$RUNNER_TEMP/appimage-smoke"
           mkdir -p "$smoke"
           cd "$smoke"
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -56,6 +56,21 @@ jobs:
         run: |
           echo "path=$(ls Copperline-*.AppImage | head -n1)" >> "$GITHUB_OUTPUT"
           echo "helper=$(ls Copperline-*-net-helper.tar.gz | head -n1)" >> "$GITHUB_OUTPUT"
+      - name: Smoke test the AppImage
+        # Boot the finished AppImage from an empty directory: the run must
+        # find the bundled AROS ROM through the AppDir's
+        # usr/share/copperline/aros lookup (romsearch.rs), not the source
+        # tree's assets/aros fallback that a repo-root cwd would satisfy.
+        # A capture run opens no window or audio device, so a clean exit
+        # plus a non-empty PNG proves the packaged image actually runs
+        # (libfuse2 for the AppImage runtime is installed above).
+        run: |
+          appimage="$PWD/${{ steps.artifact.outputs.path }}"
+          smoke="$RUNNER_TEMP/appimage-smoke"
+          mkdir -p "$smoke"
+          cd "$smoke"
+          "$appimage" --noaudio --screenshot-after 10 smoke.png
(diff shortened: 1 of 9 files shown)


LinuxJedi:
Addressed the review and the Windows failures in 91a6432e:

**Copilot comments** — all four taken: explicit `$ErrorActionPreference = "Stop"` + StrictMode in the Windows zip smoke, `chmod +x` before running the AppImage, an EXIT trap in the present smoke so a failure before `shutdown` cannot leave the emulator running until the job timeout (plus an explicit failure when the control endpoint never appears), and an EXIT trap so a failed dmg smoke cannot leave the image mounted.

**Windows test failures** — six host-portability issues, none in the emulation core (2021/2027 passed on the first-ever Windows run):

- config tests embedded temp paths in TOML *basic* strings, where `C:\Users\...` backslashes parse as escape sequences → switched to TOML literal strings.
- a filesys test compared relative paths as strings, keeping the host separator → folds `\` before comparing.
- `host_fs_usage` had no Windows implementation at all, so directory-as-HDD mounts on Windows reported a fabricated 1 GiB volume — the smoke coverage caught a real gap. Implemented via `GetDiskFreeSpaceExW` (quota-aware available bytes, matching statvfs `f_bavail`); cross-checked against the msvc target.
- the non-regular-shader-file test asserted the Unix rejection message; Windows refuses to open a directory at `File::open`, so the test now asserts the platform-appropriate rejection path.
- the CRT shader render tests ran on DX12 WARP, which renders flat grey up to 14 8-bit steps off the source — that tests WARP, not our shaders. Software (`DeviceType::Cpu`) adapters now skip like a missing adapter; real-GPU coverage stays on the macOS runners' Metal and dev machines.

Note this run is also the first time the probe goldens execute on Windows (the previous run aborted at the lib target before reaching them).

PR #355, opened 2026-07-31
launcher: boot param tidy-up 

hobbo91:
Tidy-up of the Boot Priority sub-page, so it lists what the machine can actually boot from. Visual only.

SCSI unit rows appear only once a unit carries a disk. A controller is a bus, not a drive, so fitting one no longer fills the page with seven empty rows.

A row with no drive has no stepper. The IDE rows are always listed, and read "No drive" when empty -- as text, without the [<]/[>] a priority would offer, so one that can be changed no longer looks like one that does not exist. The text starts under the "Priority" heading, where the rest of the launcher puts a greyed row's reason.

The page's screenshot in docs/guide/ui.md is regenerated to match


LinuxJedi (approved):
Another great visual cleanup. Many thanks!

Issue #50, opened 2026-06-27
Allow overriding of drive names


sidick:
Currently they're automatically created based on directory name, it could be useful sometimes to allow the overriding of this in the config and maybe GUI

PR #358, opened 2026-08-01
timing-test: real-A1200 column; model 020 posted writes and read-return sync

LinuxJedi:
## Summary

A contributor ran `timing-test.adf` on a **real A1200** (stock 68EC020 at 14.19 MHz, AGA, 2 MB chip, KS 3.2.3) - the first real-hardware column for the A1200 machine shape, where FS-UAE was previously the only 020 reference. This PR carries the disk fix that made the full table readable, records the real column (22 confidently-read rows, including a hand-typed read of rows 9-22 across both runs), and implements the two Copperline-side model corrections it demands. Two further corrections landed upstream via benletchford/m68k-rs#50 and shipped in m68k 0.5.0, which this branch now depends on: the taken-branch refill drops from 2 to 1 (real silicon: cached taken `dbra` = 7 clocks, not 8) and 020 result forwarding on register-to-register MOVE pairs (the RAW-dependent pair is one clock *faster* than the independent pair - rows 29/28: 10 vs 11 clocks).

### Test disk

- The render loop stopped at row 30, so row 31 (DIV during an active 6-bitplane display - the column's open item) was **never drawn**, and the 8-scanline pitch pushed the table bottom into CRT overscan. All 32 rows now render at a 7-scanline pitch inside the CRT-safe area, each prefixed with its two-digit decimal row ID, on a cleared bitplane. Serial output unchanged; emulated measurement rows byte-identical before/after. Golden re-blessed for the layout only.
- All five `timing-test/*.toml` configs re-pointed at `test-assets/` ROM paths (orphaned by the repo-root cleanup).

### Model changes (gated to the 020+ short-bus-cycle path; 68000 machines untouched)

- **Posted chip writes**: the 020's bus unit is decoupled from execution - a chip write is accepted after its 3-clock cycle, the transfer overlaps the following instructions (credited back against the CPU charge), and it retires into a later free chip slot at the port's 2-cck turnaround cadence; a second chip access stalls on the single in-flight cycle. Real A1200: ~8 clocks per write+dbra iteration; Copperline previously billed 11-12 (~1.5x real).
- **Read-return synchronizer clock**: +1 CPU clock per chip/custom read for the data crossing back into the CPU clock domain, accumulated into whole colour clocks. Fixes the chip-read loop (row 2, whose old exact match was the branch over-bill cancelling this) and lands the copper-vs-CPU poll row 27 on the real machine's exact beam position.
- Savestates bump to STATE_VERSION 45 (the posted-write debt and port timestamp are machine state; the m68k release also changes CpuCore's serialized shape, covered by the same bump).

### Result (with m68k 0.5.0, as on this branch)

All 22 confidently-read real rows land at 1.00: move/shift exact, the five write rows tick-exact against the hand-typed column (rows 10/12 = 415, 15 = 644), row 27 exact, row 29 at 4099 vs real 4100, row 31 at +0.9% (Copperline had this row right all along - FS-UAE over-bills it by 39%). Remaining, precisely bounded: composite write+poll rows 16/17/21 read 3-7% low, and the raw interrupt-phase rows 19/20/22 sit a few cck early of readings that themselves move between real runs - both composite effects, recorded in `docs/internals/cpu.md`.

### Testing

- 2113+ unit tests (4 new bus tests: posting/credit, drain during internal advance, back-to-back write cadence, read sync-clock accumulation) + all 25 probe goldens green on the commit state; clippy and `cargo fmt --check` clean.
- A500/68000 serial column byte-identical before/after (all new mechanisms are 020+-gated).
- TRSI Cubic Dream (issue #285) runs clean past 140 s with all three fixes.

**Merge note**: resolved - m68k 0.5.0 shipped with benletchford/m68k-rs#50 and the branch now depends on it, so the branch refill, forwarding, and these bus mechanisms (calibrated as a set) all ride in together. Full verification re-run against the released crate: 2113 unit tests, 25 probe goldens, the real-A1200 comparison (every confirmed row at ratio 1.00), clippy/fmt, and save/resume byte-determinism.


LinuxJedi:
To be merged after the m68k-rs patch is in a release

LinuxJedi:
This is merged but GitHub got stuck

Issue #371, opened 2026-08-03
Sanity Roots 2 has lots of AGA issues with A1200


LinuxJedi:
One I haven't tested for a couple of releases, but there are numerous issues with it now.

Issue #60, opened 2026-06-28
[Bug]: keyboard doesn't work


sharpied79:
### Copperline version

0.8.0

### Host system

Fedora 44 (KDE Plasma)

### Machine/configuration

```toml
All
```

### Reproduction steps

Keyboard doesn't appear to work?

### Expected hardware behaviour

Keyboard should work

### Actual behaviour

Keyboard doesn't work

### Logs or debugger output

```text

```

LinuxJedi:
@sharpied79 have you changed "Joystick Input" to "Gamepad"? The joystick controls take over some of the keyboard by default if you don't have a USB joystick plugged in (0.9.0 will have a boot-time config option for this).

sharpied79:
Ah yes, that sorted it, just had a "duhh" moment sorry.

LinuxJedi:
No problem, I have an idea on to make that more visible, I'll try and flesh it out.

LinuxJedi:
@sharpied79 behaviour change in the PR for 0.9.0:

1. The "auto" mode is going away. "gamepad" will be the default.
2. There is a toggle button on the status bar, it will show a gamepad for gamepad mode and keyboard for keyboard controls.

Hopefully this will help in the future.

PR #373, opened 2026-08-03
denise: fold wide-FMODE scroll taps at the absolute-grid arrival phase (issue #371)

LinuxJedi:
## Summary

Fixes the two visible regressions SANITY Roots II AGA picked up in v0.13 (issue #371, regressed by #250): the swirl screen sitting one full gulp left and the kaleidoscope screen shearing line by line, plus the perceived slowdown those caused.

**The corrected hardware rule:** Denise's reload comparator does not know where the bitplane fetch started - it matches the BPLCON1 tap against a free-running per-line pixel counter, an absolute gulp grid anchored at hpos 0 (the WinUAE cycle-exact `delay_cycles` model). A wide-FMODE fetch with an off-grid DDFSTRT arrives early, so the fold boundary is the arrival distance: **earliness + the 8-cck fetch-to-comparator pipeline, saturating** (no wrap) - once the boundary passes the top of the tap range nothing folds, and an exactly on-grid start folds from the pipeline alone.

PR #250's last-earliness-window rule (`gulp - earliness`) matched Alien Breed II AGA only by coincidence (earliness 8 px on 32-px gulps puts both formulas at 24) and folded every Roots II tap >= 16 where the true boundary sits past the whole 0..63 range.

## Verification

- New `ddfprobe-agafold2` golden probe sweeps the DDFSTRT phase (0/4/8/16/24/28 cck) against taps across the boundary on the 64-bit fetch. **FS-UAE renders all 16 bands identically** - including the saturating phases 24/28 and the on-grid fold at tap 56 that both the wrapped and the last-earliness models miss. The FS-UAE relative band map is recorded in the probe header.
- The `ddfprobe-agafold` (AB2) golden is **bit-identical** - both rules agree on that constellation - and kicka1200 / cd32 / zool screenshots are byte-identical to the pre-fix build.
- Roots II fresh boot: the swirl screen is centred again and the kaleidoscope is clean, matching v0.12; the 25/35/60/100s scenes are byte-identical to before.
- Full suite green (2224 tests incl. all 27 goldens), clippy and fmt clean.
- Hi-res/SHRES pipeline scaling remains unverified (TODO in code), as it was in #250.

## Also included

A new `bfprobe` timing-test disk (14 rows) for the 68020 bit-field instruction class, which the same issue's "DIE" dissolve scene exposed: its per-frame worker plots two `BFSET (An){Dn:1}` per pixel under a hard one-frame deadline. The scene's slowdown itself is fixed in the m68k crate - see benletchford/m68k-rs#60 (five-byte-burst window reads plus double-billed MC68020UM operand cycles); Copperline picks that up with the dep bump once it releases. No real-hardware column exists for the bit-field class anywhere, and the Copperline and FS-UAE columns recorded in the probe header already disagree structurally on multi-byte spans - a real-A1200 photo of this disk would settle it.

Part of #371; the ~135s copper-sprite rotozoom scene remains open separately.


dlundqvist (commented):
From my A1200 (if you need): Image

Issue #366, opened 2026-08-02
CI: Windows unit-test binary died once with STATUS_ACCESS_VIOLATION


LinuxJedi:
Filing this to have somewhere to collect occurrences. Seen once, on CI, not reproduced since.

## What happened

The Windows `Unit tests` job died mid-run with an access violation instead of failing a test:

```
test video::window::tests::autofire_does_not_gate_scripted_joystick_input ... ok
test video::window::tests::bar_layout_stacks_three_or_more_drives_two_up ... ok
Exception: D:\a\_temp\...ps1:4
   4 |  … ($LASTEXITCODE -ne 0) { throw "unit-lib exited with $LASTEXITCODE" }
     | unit-lib exited with -1073741819
```

`-1073741819` is `0xC0000005`, `STATUS_ACCESS_VIOLATION`. No panic, no test failure, no `test result:` line: the process was killed outright.

Run: attempt 1 of https://github.com/CopperlineHQ/Copperline/actions/runs/30762526174 (job `Unit tests`), commit f1eb5500 on the branch behind #365. windows-latest, `x86_64-pc-windows-msvc`, running the prebuilt release `ci-bins/unit-lib.exe` staged by the `test-bins` job.

The harness announced `running 2146 tests` and 1881 of them had reported ok when it died, about 6.2 seconds in (19:16:29.0 start, 19:16:35.2 crash). Re-running the same job on the same commit passed in 10.1s (2136 passed, 0 failed, 10 ignored), and the previous twelve runs of that workflow were all green, so it is neither reproducible nor an established flake.

The last completions before the crash are in `video::window::tests` and `serial::tests`, but the harness runs tests in parallel, so that dates the crash rather than naming the culprit. The 265 that had not reported yet are the interesting set and I have no way to narrow it from this log.

## Guesses, in rough order

- One of the GPU-backed suites (`video::window::crt_shader`, `video::window::rtg_texture`, and the `pixels`-backed window tests). windows-latest has no GPU, so those run against WARP/software D3D12; a fault inside the driver, or a device torn down while another thread still holds it, looks exactly like this and would be timing-dependent enough to show up once in a dozen runs.
- A test that leaves a thread running past its own end (audio sinks, serial, filesys workers have such threads) and touches state the harness has already dropped. `serial::tests::tcp_connect_to_nothing_is_an_error` completing next to the crash is at least suggestive.
[...]

PR #378, opened 2026-08-03
window: serialize the offscreen GPU tests on one shared device helper (issue #366)

LinuxJedi:
## Summary

Fixes the intermittent Windows CI crash tracked in #366: the `Unit tests` job has died three times with `STATUS_ACCESS_VIOLATION` (0xC0000005), most recently on the merges of #367 and #376. Reconstructing the in-flight test set from the crashed logs (full analysis in [this comment](https://github.com/CopperlineHQ/Copperline/issues/366#issuecomment-5171643146)) shows every crash died with the bezel/rtg_texture offscreen render tests running concurrently on separate threads, the two merge-commit crashes at the identical schedule point after a 6-7s stall.

The root cause those sets point at: `crt_shader::gpu()` already refuses software rasterizers because WARP does not render the passes faithfully, but the bezel and rtg_texture copies of the helper never picked up that skip. They were therefore the only tests creating D3D12 WARP devices on the runners - up to three concurrent instance/device create-render-teardown cycles at once, which is exactly the kind of load that takes WARP down.

## Changes

- The three per-suite `gpu()` helpers become one shared `crt_shader::test_gpu(label)`, next to the `poll_once` they already shared.
- A process-wide mutex serializes the GPU tests: the guard is returned with the device and held for the life of the test, so concurrent device lifecycle on one adapter is gone entirely.
- The software-rasterizer skip now applies to all three suites, so windows-latest creates no WARP devices at all - matching the policy the crt_shader suite already established. The render tests still run for real on the macOS runners' Metal GPU and on developer machines.

Test-only change; no behavior change in the emulator or the shipped binaries.

## Testing

- `cargo test` clean (2153 passed; the offscreen render tests exercised for real against Metal).
- `cargo clippy --all-targets` and `cargo fmt --check` clean.

### DIFF
--- a/src/video/window/bezel.rs
+++ b/src/video/window/bezel.rs
@@ -341,26 +341,10 @@ mod tests {
     /// left a hole, and any outside means it overdrew its viewport.
     const SENTINEL: [u8; 4] = [0, 0, 255, 255];
 
-    /// A device, or `None` on a machine with no usable adapter (headless
-    /// CI): the render tests then pass without asserting anything.
-    fn gpu() -> Option<(wgpu::Device, wgpu::Queue)> {
-        let instance = wgpu::Instance::new(wgpu::InstanceDescriptor::new_without_display_handle());
-        let adapter =
-            match crt_shader::poll_once(instance.request_adapter(&wgpu::RequestAdapterOptions {
-                power_preference: wgpu::PowerPreference::LowPower,
-                force_fallback_adapter: false,
-                compatible_surface: None,
-            })) {
-                Some(Ok(adapter)) => adapter,
-                _ => return None,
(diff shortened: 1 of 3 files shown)


LinuxJedi:
Addressed the review in 4226e118:

- `test_gpu` now returns a `TestGpu` handle with the lock in a private field and the device/queue borrowed through accessors, so a caller can no longer destructure the guard away and reintroduce concurrent device lifecycles.
- The skip messages moved into the helper, which knows the actual reason: "no GPU adapter", "software rasterizer (name)", or a failed device request. The call sites no longer print a misleading "no GPU adapter" for the WARP skip.

PR #391, opened 2026-08-05
z3660: honour the minterm on a rect copy

codewiz:
`OP_COPYRECT_NOMASK` carries a minterm in the same low-nibble form the planar
hooks use, but it was compared against `0xC0`, the byte form:

```rust
        const MINTERM_SRC_IDX: u8 = 12;   /* used by the planar paths */
        const MINTERM_SRC: u8 = 0xC0;     /* used here */
...
                if op == OP_COPYRECT_NOMASK && minterm != MINTERM_SRC {
                    log::debug!("z3660: copyrect minterm {minterm:#04x} treated as SRC");
                }
```

Nothing ever matched, so every rect copy took the "treated as SRC" fallback.
Plain copies came out right by luck; a clear became a copy of whatever sat at
the source. Under AROS that shows up as windows drawing the desktop behind
them, because its layer backfill clears a region with a rect copy at minterm
0 rather than with a fill.

Compare against `MINTERM_SRC_IDX` like the planar paths do, and put anything
else through the `minterm_apply()` helper that already exists a few lines up.
The plain-copy path keeps its byte-at-a-time loop; only the minterm case goes
pixel by pixel.

The other emulated board agrees on the index form -- amiberry's ZZ9000 does
`switch (minterm & 15)` with `ZZ_MINTERM_SRC = 12` -- as do the ZZ9000 and
Z3660 firmwares and AROS's p96gfx.

Tested with the p96cts conformance suite on AROS/m68k with a Z3660 board:
`ScrollRaster-backfill` failed there and nowhere else, and passes with this.
A scroll backfills the region it vacates, which is the same clear. The run
also brought `BltBitMap-friend` on that board from 12615 differing pixels to
1872, matching what ZZ9000 already reported.

### DIFF
--- a/src/z3660.rs
+++ b/src/z3660.rs
@@ -389,7 +389,6 @@ impl Z3660 {
         const COMPLEMENT: u8 = 2;
         const OP_SPRITE_COLOR: u32 = 12;
         const OP_SPRITE_BITMAP: u32 = 13;
-        const MINTERM_SRC: u8 = 0xC0;
         const JAM1: u8 = 0;
         // Drawmode bits (firmware rtg/gfx.h): JAM1/JAM2 in bit 0, COMPLEMENT
         // in bit 1, INVERSVID in bit 2.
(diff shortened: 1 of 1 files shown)


codewiz:
Before this fix, the AROS console would appear transparent in Z3660 8bpp modes:
Screenshot_20260805_061212

LinuxJedi (approved):

PR #394, opened 2026-08-05
Replace vendored FloppyBridge with FluxBridge (mk2) 

hobbo91:
## Summary

- replace the vendored C++ FloppyDriveBridge runtime and C ABI wrapper with the safe Rust FluxBridge v0.3.0 crate, tracking its main branch and compiled with the Greaseweazle driver only
- serve a track's first read while the platter is still turning it, so the guest reads behind the head as a real controller does; boots reach the Workbench desktop faster than the plugin did
- trust captures by FluxBridge's own quality verdict (join proof plus AmigaDOS scan); Copperline's duplicate revolution scanner is deleted
- build the launcher's interface list and the Configure page's library name and version from the crate itself; no driver knowledge is hardcoded, and adding a future driver is a Cargo feature
- simplify the surface: read modes are normal/compatible/stalling, replay_speed is fast (default) or normal, auto-cache is gone, and density auto is genuinely sensed from the flux
- report media by the drive's own state: insert/eject on the OSD (with the disk's tab state on a writable drive) and in the log, with COPPERLINE_DIAG_FLUXBRIDGE for the full commentary
- remove the C++ build dependency and the vendored source tree; disk-image bays are untouched
- rewrite the physical-drive guide and update requirements, example config, launcher labels, and credits

FluxBridge grew from a Rust port of Rob Smith's FloppyDriveBridge and carries the original-project attribution and license provenance in its README, NOTICE, and docs.

## Validation

- cargo fmt --all -- --check
- cargo test (2,204 library tests plus binary and integration tests; the one failure, cpu::tests::three_byte_operand_moves_exactly_three_bytes, fails identically on main and is unrelated)
- cargo clippy --all-targets -- -D warnings
- cargo check --no-default-features, and --no-default-features --features frontend
- hardware: repeated headless Workbench 1.3 boots to the desktop from a physical disk (Greaseweazle V4.1, A500, Kickstart 1.3), screenshot-verified; all three read modes, both replay tiers, and the port, cable, and density options exercised against the drive
- physical writes tested from the guest against a real disk and verified working
- verified end to end on macOS, Windows 11, and Linux hosts


src/fluxbridge/mod.rs
```diff
@@ -0,0 +1,834 @@
...
+        let cylinder = self.clamp_cylinder(cylinder);
+        self.side = side_of(side);
+        self.cylinder = cylinder;
+        let request = fb::WriteRequest {
+            track: self.track(),
+            words: words.to_vec(),
+            bit_len: words.len() * 16,
+            start_bit,
+        };
+        match self.inner.submit_write(request) {
```
LinuxJedi, line 530: [P1] Handle async write completion before recaching the track

submit_write only queues the physical write. The pinned FluxBridge leaves the prior capture readable until its worker emits WriteCompleted or WriteFailed. finish_write_words invalidates Copperline's cache immediately, so an intervening read can fetch the stale pre-write capture and promote it back to BridgeTrack::Kept. No caller currently polls take_write_failure, and that method ignores WriteCompleted, so the stale data can persist indefinitely and late failures remain silent. Please track pending writes, prevent old captures from being recached, and invalidate/re-read the affected track when completion or failure arrives.

src/floppy.rs
```diff
@@ -2166,6 +2200,32 @@ impl FloppyController {
...
+                // the head spotted this much of a start.
+                if let Some((words, bits)) = bridge.partial_track(cylinder, side) {
+                    if bits >= BRIDGE_PARTIAL_MIN_BITS {
+                        let nominal = encoded_track_words();
+                        drive.cached.revs = vec![TrackRev::new(
+                            words,
+                            bits,
+                            Self::word_cck_for_track_words(nominal),
+                        )];
+                        drive.bridge_filler_track = None;
```
LinuxJedi, line 2281: [P2] Retain partial captures between polling ticks

This installs the partial snapshot but clears bridge_filler_track without setting cached_track or another retention marker. On the next emulator tick, ensure_track misses the current-cache fast path, sees keep_filler as false, and clears drive.cached before the one-millisecond polling deadline is reached. The partial is therefore visible for only one device tick instead of continuously until the next snapshot. Please retain the partial capture until it is refreshed or replaced by the completed revolution.

Cargo.toml
```diff
@@ -148,6 +143,15 @@ smoltcp = { version = "0.12", default-features = false, features = [
...
 zerocopy = { version = "0.8", features = ["derive"] }
+# Physical floppy drives: CopperlineHQ/FluxBridge, tracking its main branch.
+# Pure Rust -- the interfaces sit behind one typed API with no C or C++ ABI,
+# so nothing has to be installed and nothing is vendored. `Cargo.lock` still
+# fixes the exact commit every build uses, so a checkout is reproducible;
+# `cargo update -p fluxbridge` is how a newer one is taken. Only the drivers
+# named here are compiled in, and the launcher's interface list follows this
+# line: adding a driver Copperline supports means adding its feature, nothing
+# else.
+fluxbridge = { git = "https://github.com/CopperlineHQ/FluxBridge", branch = "main", default-features = false, features = ["greaseweazle"], optional = true }
```
LinuxJedi, line 154: [P2] Advertise only the drivers compiled into this build

Only the Greaseweazle FluxBridge feature is compiled here, but --help and configuration validation still advertise DrawBridge and SuperCard Pro. Selecting either reaches attach_floppy_bridges and fails because driver_named returns None. Please derive the advertised/accepted choices from fluxbridge::drivers(), or restrict them to Greaseweazle for this build.

docs/guide/configuration.md
```diff
@@ -53,8 +53,7 @@ range checks as the equivalent TOML fields:
 | `--floppy-bridge-cable DFN SEL` | `[floppy.dfN] bridge_cable` | drive select: `a`/`b` (IBM PC cable) or `0`-`3` (Shugart) |
 | `--floppy-bridge-mode DFN MODE` | `[floppy.dfN] bridge_mode` | how tracks are captured: `normal`, `compatible`, `stalling` |
 | `--floppy-bridge-density DFN D` | `[floppy.dfN] bridge_density` | force a density: `auto`, `dd`, `hd` |
-| `--floppy-bridge-speed DFN PCT` | `[floppy.dfN] bridge_speed` | serve captured tracks at `100`, `125`, `150`, `175`, or `200` percent of real speed |
-| `--floppy-bridge-auto-cache DFN` | `[floppy.dfN] bridge_auto_cache = true` | cache disk data while the drive is idle |
+| `--floppy-replay-speed DFN PCT` | `[floppy.dfN] replay_speed` | serve captured tracks at `100`, `125`, `150`, `175`, or `200` percent of real speed |
```
LinuxJedi: [P2] List only accepted replay-speed values

This row says 125, 150, and 175 are valid, but both the CLI and TOML parser accept only normal/fast or their 100/200 aliases. Commands copied from this guide therefore fail validation. Please list only the values the parsers accept, in line with the repository's CLI-documentation requirement.

LinuxJedi (changes requested):
Codex review: This change has a blocking data-correctness problem in the asynchronous write path, and the partial-capture cache currently drops streamed data between polling ticks. The advertised physical-drive and replay-speed choices also need to match what this build and its parsers actually accept. Details are inline.

src/fluxbridge/mod.rs
hobbo91: Confirmed and fixed in `46587a68e39c0aed57a9e8206b3a69ac05aa448f` — this was real, and worse than a stale read.

`take_write_failure` was never called from anywhere, so the events were never drained at all. Combined with dropping the cache at submit time, a read landing between submit and completion fetched FluxBridge's pre-write capture, and if that capture proved out it was filed as `BridgeTrack::Kept` and served from memory indefinitely — no bound on how long the guest saw a disk that no longer existed. Failures were silent for the same reason.

Now: `bridge_writing` holds the track from the moment the write is accepted, and `ensure_track` refuses to read or cache it while pending, so no stale capture can be promoted. The adapter's `take_write_failure` is replaced by `poll_write_outcomes`, which reports **completions as well as failures** — completion is what says the track may be read again — and it is drained every frame from `poll_bridge_media`. Both outcomes release the track and reset it to `BridgeTrack::Unknown`, since either way the platter holds something the guest has not seen; a failure now logs a warning naming the track. Pending writes are also cleared when the disk changes or the drive is released, where no outcome will ever arrive. Regression test: `a_track_being_written_is_not_read_from_the_drive`.

src/floppy.rs
hobbo91: Confirmed and fixed. Traced it exactly as you describe: the partial set `cached.revs` but no retention marker, so the next tick took `keep_filler == false` and cleared `drive.cached` at the top of `ensure_track`, leaving the head with nothing until the poll deadline expired — the partial was live for one tick out of the whole interval.

Added `bridge_partial_track`, and the retention test is now `bridge_filler_track == Some(track) || bridge_partial_track == Some(track)` (renamed to `keep_provisional`, since filler and a part-captured revolution are both provisional data for the same track). It is cleared when the finished revolution supersedes it, when a write is submitted, and on disk change or release. Regression test: `a_partial_capture_is_retained_between_ticks`.

Cargo.toml
hobbo91: Agreed — fixed by deriving it, per your first suggestion.

Added `config::supported_bridge_drivers()`, which reads FluxBridge's own driver table, and it now feeds the CLI help, the `--floppy-bridge` usage line, and the parser's unknown-name error. A recognised name whose driver is not compiled in is refused at parse time naming what *is* available (`this build has no DrawBridge driver (it has greaseweazle, off)`), rather than validating and then failing at `attach_floppy_bridges` with the machine half built. The launcher's interface row was already built from `drivers()`, so all four surfaces now agree.

New test `a_driver_this_build_lacks_is_refused_with_what_it_has`, and it immediately caught an existing test that configured DrawBridge on df1 — updated.

docs/guide/configuration.md
hobbo91: Fixed. That row was left behind when the ladder was retired: it now reads `fast` (the default, double speed) or `normal` (the platter's own), matching both parsers, and the flag name is corrected to `SPEED` rather than `PCT`.

While there I found the row above it had the same problem — it advertised `drawbridge`, `supercardpro` — so that one now says `greaseweazle`, or `off`, in line with the driver-table change in the Cargo.toml thread.

hobbo91:
All four addressed in `46587a6`, replies on each thread.

The P1 was real and I'd underrated it: `take_write_failure` was never called from anywhere, so no write event was ever drained. A track written by the guest could be re-read from FluxBridge's pre-write capture, proved out, and then filed as faithful — served from memory indefinitely while the platter held something else. Writes that failed on the platter were silent for the same reason. A track is now held unreadable from the moment its write is accepted until the drive reports the outcome, completions included, and a failure warns naming the track.

The P2s were both as described: the partial capture had no retention marker so it survived a single tick, and the CLI/parser/docs advertised three interfaces where the build compiles one. The driver list is now derived from FluxBridge's own table in all four places, so it cannot drift again.

Three new regression tests. Full suite, clippy and fmt clean — the two failures on this branch (`cpu::three_byte_operand`, and `bar_layout` under the parallel suite only) reproduce identically on upstream main and are not from this PR.

Worth flagging separately: Copilot declined to review this PR because it exceeds 20,000 changed lines. It is 20,538 — of which 16,435 are the vendored C++ tree being deleted. The authored change is ~1,800 lines.

LinuxJedi (approved):
Fantastic work! Many thanks for this!

PR #402, opened 2026-08-06
web: rebuild the audio stack when iOS kills the page's audio session

LinuxJedi:
## Problem

On iOS, switching from Safari to another app and back loses audio permanently. The stat line still shows a live, low audio buffer -- the worklet's render thread is running and draining the queue -- but nothing reaches the speaker.

Backgrounding the browser on iOS deactivates the page's OS-level audio session. On return, WebKit can reconnect the AudioContext's render thread without reactivating the session, and the context *reports* `running` -- which makes the show-path `resume()` a spec no-op, so the page had no way back. Two sibling failure modes share the trigger: the context can come back in WebKit's non-standard `interrupted` state with `resume()` refused outside a gesture, and a phone call can reroute the session onto reconfigured output hardware.

## Fix

- Extract the boot-time AudioContext + worklet construction into `buildAudioStack()`, generation-guarded so overlapping builds settle on the newest stack.
- Add `recoverAudio()`, run on every return to the foreground (`visibilitychange` show-path, plus bfcache `pageshow` restores, which the visibility handler never sees):
  - **iOS WebKit** (detected incl. iPadOS masquerading as macOS): rebuild the stack outright -- a fresh context is the one move that reactivates the session in all three failure modes, and the discarded queue holds pre-switch audio nobody missed. If the return does not count as a gesture, the existing unlock-on-next-input listeners hand the job to the next tap.
  - **Elsewhere**: plain `resume()` as before, but the worklet's ~29 ms queue reports now double as a liveness check -- a context still silent 600 ms after its resume gets the same rebuild.
- A rebuild while paused leaves the fresh context suspended, keeping the pause contract. This also fixes a pre-existing blip: the old show-path resumed a paused machine's context on tab return, replaying stale queued audio.
- `window.__audioCtx` is exposed alongside `window.__emu` for automation.

Documented in `docs/guide/browser.md`.

## Testing

Staged the site with this glue in a driven Chrome session (module-scoped, so verified via `window.__audioCtx` identity + state):

- Desktop path: hide suspends, show resumes the *same* context; heartbeat check does not fire a spurious rebuild.
- Heartbeat path: with `resume()` sabotaged to a no-op, the silent context is detected and rebuilt within the check window; machine keeps stepping.
- iOS path (detection forced on in the staged copy): every foreground return closes the old context and brings up a fresh running one; queue refills; 50+ fps throughout.
- Generation guard: rapid hidden/visible flip-flops race two rebuilds and settle on one live running context.
- Pause contract: rebuild-while-paused stays suspended; unpause resumes it.

Not verifiable from here on real iOS hardware -- needs a quick confirmation on an actual iPhone/iPad that returning from an app switch brings sound back.

The website repo carries a copy of this glue until the next tag republish; companion PR: CopperlineHQ/copperline.github.io (same try.js).

### DIFF
--- a/crates/copperline-web/www/try.js
+++ b/crates/copperline-web/www/try.js
@@ -488,10 +488,15 @@ async function load() {
 // closes the previous stack first so it cannot keep playing alongside
 // the new one (a reboot after an emulator error, an iOS audio-session
 // revival); the generation counter makes overlapping builds settle on
-// the newest one.
+// the newest one. suspendForPause is the caller's intent for a machine
+// that is paused right now: a foreground recovery passes the live
+// paused flag to keep the pause contract that audio stays suspended,
+// while boot always builds a running stack - it starts the new machine
+// unpaused, and clears the paused flag itself without going through
+// the unpause path that would otherwise resume the context.
 let audioBuildGeneration = 0;
 
-async function buildAudioStack() {
+async function buildAudioStack(suspendForPause) {
   const gen = ++audioBuildGeneration;
   if (audioCtx) {
     audioNode?.disconnect();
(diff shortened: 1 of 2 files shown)


LinuxJedi:
Both Copilot findings were real and are addressed in e7a8005:

1. **Reboot from pause ran silent** -- confirmed: `boot()` clears the paused flag directly, and the machine/video selects call `boot()` on a paused-but-running machine, so the fresh context got suspended with nothing left to resume it. `buildAudioStack()` now takes the caller's pause intent explicitly: boot always builds a running stack, foreground recovery passes the live flag.

2. **A transient rebuild failure silenced the session for good** -- confirmed, and it went one step deeper than the comment: besides `recoverAudio()`'s null guard, the `visibilitychange` handler's own top-level `if (!audioCtx) return` also kept the retry from ever running. The failure path now closes the half-built context and rethrows, the show path reaches `recoverAudio()` with no stack installed, and a running machine gets one logged rebuild retry per foreground return.

Both verified in the driven-Chrome harness: a machine-model change while paused now boots with running audio, and a rebuild sabotaged to throw at worklet construction leaves the machine stepping and recovers a running stack on the next foreground return.

PR #403, opened 2026-08-07
web: unlock audio on pointerup, and keep the unlock armed until it lands

LinuxJedi:
Follow-up to #402, which rebuilt the audio stack on returning to the foreground but still left a real iPhone silent after an app switch. On-device Web Inspector data pinned the true mechanism, and it was in our unlock listener all along:

## Root cause

- After the return, the rebuilt context sat `suspended` with `currentTime: 0` -- built fine, never started, its gestureless `resume()` refused. That part is by design; the tap was supposed to finish the job.
- **The tap could not work.** User activation on a touch screen is granted at the *end* of the gesture: the HTML spec's activation-triggering events count `pointerdown` only for mice, while a finger's activation arrives at `pointerup`/`touchend`. Our unlock listened on `pointerdown`, so it always fired *before* the tap carried activation, iOS refused the resume -- and the listener had already removed itself, so the first tap permanently burned the unlock.
- Desktop testing could never catch this: a mouse's `pointerdown` **does** grant activation.
- A page reload cures it because the boot click creates the context inside a full activation window -- confirmed on-device.

## Fix

- Unlock triggers are now `pointerup` + `keydown`; by pointerup both touch and mouse gestures hold activation.
- The listeners stay armed until a resume verifiably reaches `running`; a refused attempt no longer costs the arming.
- Escalation ladder: if an in-gesture resume settles with the context still stuck, the next gesture rebuilds the whole stack *inside* its activation window -- the same shape as the boot click the reload evidence proved reliable. Any fresh build resets the ladder.

## Testing

Driven-Chrome harness from a page with no user activation at all (the iPhone's post-interruption state): `pointerdown` is inert; a refused in-gesture resume stays armed and escalates the next `pointerup` into a full rebuild; a landing resume disarms both triggers (resume-call counter reads zero on subsequent fires); the foreground-return rebuild and all #402 flows still pass. The activation grant itself cannot be exercised in the harness (an occluded automation tab receives no real input) -- it rests on the spec semantics and needs the usual iPhone confirmation after the site republish.

Expected behavior on the phone after this lands: switch away, switch back, **first tap brings the sound back** (a second tap covers the pathological case via the rebuild rung).

### DIFF
--- a/crates/copperline-web/www/try.js
+++ b/crates/copperline-web/www/try.js
@@ -523,6 +523,14 @@ let audioBuildGeneration = 0;
 let audioUnlockRebuild = false;
 
 function audioUnlock() {
+  // The pause contract holds on this path too: a paused machine's
+  // context stays suspended, and unpausing owns the resume. Returning
+  // WITHOUT disarming keeps the ladder ready for the first unpaused
+  // gesture - which cannot be the unpause tap itself (its pointerup
+  // precedes the click that clears the flag), but setPaused resumes
+  // inside that same activation window, and a resume it cannot land is
+  // exactly what the still-armed ladder is for.
+  if (paused) return;
   if (!audioCtx || audioUnlockRebuild) {
     audioUnlockRebuild = false;
     buildAudioStack(false).catch((e) => console.error('audio rebuild', e));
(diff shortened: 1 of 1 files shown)


LinuxJedi:
Valid catch, addressed in a9a394f9 -- and it went one step further than the comment: besides the paused-gesture hole, the race it describes (the Pause tap's own pointerup firing the unlock while still unpaused, with the click's suspend landing before the resume settles) was *seeding the escalation flag*, so the first gesture after unpause did a needless full rebuild. Now:

- `audioUnlock` returns early while paused, **without disarming**, so the ladder stays ready for the first unpaused gesture; the unpause tap itself resumes through `setPaused` inside the same activation window, and a resume it cannot land is exactly what the still-armed ladder covers.
- A suspended reading (or rejection) that settles while paused no longer arms the escalation -- that is the pause contract at work, not a context beyond resuming.

Harness-verified: gestures while paused call no resume and rebuild nothing; the settle-after-pause race leaves escalation unarmed (first unpaused gesture takes the plain-resume rung on the same context); a landing resume still disarms both triggers.

PR #418, opened 2026-08-08
Direct WHDLoad boot: --whdload, launcher integration, packaged support archives

### LINKED ISSUE #414: WHDLoad support
I believe other emulators allow for direct WHDLoad support. We should look into that.

LinuxJedi:
Closes #414.

Boots a WHDLoad-installed game package directly, with no Workbench disk or hand-built hard drive:

```sh
copperline --whdload Turrican.lha
```

Also available as `[whdload]` in the configuration, a **Storage -> WHDLoad** launcher page, and `.lha`/`.slave` drag-and-drop (running machine reboots into the game; the configuration screen fills its game field).

## How it works

- **Two live hostfs mounts, no disk images.** The package is extracted once into a per-game library directory (`WHDGame:`) and reused on later runs, so savegames persist on the host; a boot volume (`WHDBoot:`, bootpri 6) is regenerated each launch around the real WHDLoad binary: `C/WHDLoad`, a generated `Startup-Sequence` (`WHDLoad <slave> Preload SplashDelay=0` + `[whdload] args`), and `Devs/Kickstarts/`. Passing a directory instead of an archive mounts it in place.
- **Machine derived from the slave header** (new parser for the WHDLoadSlave structure incl. the 16.1 multi-Kickstart table): the canonical WHDLoad host, an A1200 with 8 MiB fast RAM, which satisfies `WHDLF_Req68020` and `WHDLF_ReqAGA`. Explicit `[machine]`/`rom`/`[memory]`/CLI choices always win, so `--whdload game.lha --model A4000` does what it says.
- **Kickstart images identified by content, not filename**, using the ANSI CRC-16 WHDLoad itself uses, after undoing byte-swapped dumps, doubled 256 KiB images, and Cloanto `rom.key` encryption. Identified images are staged under their canonical `Devs:Kickstarts/` names with the matching skick `.RTB` tables; a staged `kick40068.A1200` doubles as the machine ROM (bundled AROS as reduced-compatibility fallback). A slave demanding a missing image fails naming the exact file, size, and CRC it wants.
- **LHA support** lands as `src/lha.rs` over the pure-Rust `delharc` crate (member CRCs verified, hostile member paths rejected).

## Support archives: assembled at packaging time, never committed

`WHDLoad_usr.lha` (freeware since 18.2, shipped unmodified from whdload.de) and `skick346.lha` (Aminet) are fetched with pinned SHA256s by `tools/fetch-whdload.sh` into gitignored `assets/whdboot/`. The macOS/AppImage builds run the script, the Windows build fetches in PowerShell with the same pins, the Flatpak manifest declares them as pinned file sources, and the Homebrew formula stages them as `:nounzip` resources. whdload.de's file is a rolling latest-release, so an upstream WHDLoad release intentionally fails the checksum until reviewed; the fetch script header lists every pin site. Kickstart ROMs remain the user's own and are never shipped.

## Testing

- New unit coverage: LHA round-trips against hand-built lh0 fixtures, slave/multi-kick parsing (cross-checked against the assembled fixture slave), Kickstart dump-form normalization, staging, config derivation, launcher round-trips (incl. `args` passthrough), drop classification. Full suite 2289 green; `clippy --all-targets --all-features -D warnings` and `fmt --check` clean.
- `tests/whdload_boot.rs`: asset-gated end-to-end boot of the committed, project-owned fixture package (`tests/assets/whdload/`, a slave that paints COLOR00 $0B4 and parks) through the real WHDLoad binary under a local Kickstart 3.1, asserting the solid frame; skips cleanly without the fetched archives or a content-identified 40.068 A1200 image.
- Verified against a real ROM collection: mislabelled/doubled/A600-revision dumps all identified correctly by content.

## Not in v1 (documented)

Per-game settings database (Amiberry `whdload_db.xml`), `.lzx` archives, icon-tooltype options, browser build support.


LinuxJedi:
CI and review findings addressed:

- **Lockfiles / flatpak sources** (`64e23bac`): the web crate's own `Cargo.lock` and `packaging/flatpak/cargo-sources.json` are regenerated for the new `delharc` dependency; the wasm32 `--locked` check passes locally again.
- **All five review findings** (`be892431`), each with a regression test:
  - slave-declared Kickstart names are validated as bare file names at parse time, closing the `Devs:Kickstarts/` staging path traversal;
  - declared requirements now only count as met by an exact size+CRC content match, staged under the declared name even over a canonical collision, so a mismatch fails host-side with the precise ask instead of later inside the guest;
  - the slave search inspects `DirEntry::file_type()` and skips symlinks, so a link to an ancestor cannot recurse;
  - extraction is atomic (temp sibling promoted by rename) and reuse requires a completion marker recording the source archive's name -- interrupted extractions are redone, same-stem collisions between different archives are refused by name;
  - the CLI boot derives the machine on a clone of the raw config and keeps the user's own settings (plus `[whdload] game`) for the session, matching the launcher's Run path, so reopening/saving the launcher never persists the derived machine or mounts and re-running cannot duplicate them.

PR #419, opened 2026-08-08
Feature/image toolkit

hobbo91:
## Create Image: making fresh ADF and HDF images

Copperline could open disk images but not make one, so a blank ADF or a
partitioned HDF had to come from another tool. **Storage -> Create Image...**
now makes both.

`src/diskimage.rs` is pure `std` with no dependencies. Two independent
sources pinned the layouts down and agree: amitools, which states the block
structures directly, and WinUAE's `disk.cpp`, which writes the same bytes
with the constants folded in. Where WinUAE has a magic number, the
derivation is noted beside it.

Nothing here touches the emulated machine. The workshop lives on the
launcher state rather than in `MachineSetup`, so none of it reaches a
config file; a test works every control on all three pages and compares
the saved TOML to prove it.

### Floppy Disk

DD or HD, standard or `UAE-1ADF` extended container, unformatted or any of
the eight DOS types, volume name, and boot code.

### Hard Disk

Size to 9999 MB/GB, RDB or no partition table, filesystem, device and
volume name, bootable with a `de_BootPri`, host read-only, and sparse (on
by default). A **Configure** page behind Custom geometry sets cylinders,
surfaces, sectors and reserved blocks, plus the drive identity.

### Filesystem and DOSType

The picker was a list of eight tags walked with arrows, which is the tag's
encoding rather than the choice being made. It is now two rows of ticks —
a family, and the options that family carries — with the rules taken from
amitools' own predicates:

    is_intl(dt) = is_dircache(dt) or is_longname(dt) or (dt & 2)

Dir cache and Long names are two values of one two-bit field, so each greys
the other; both are international, so International shows ticked and greyed
alongside either. Three boxes have eight states, the field has four, and the
test clicks every box in every order from every tag to check the four the
page produces are the four that exist.

### RDB drive identity

HDToolBox showed a drive called `Copperli` of type `ne`. The identity is a
SCSI INQUIRY reply — 8-byte vendor, 16-byte product, 4-byte revision — and
both places that wrote it treated the first two as one 24-byte field, so any
name past eight characters split across the columns. The synthesized RDB did
the same to the mounted filename, so `workbench.hdf` came up as `workbenc` /
`h.hdf`. One writer now, in `harddrive.rs`. A drive names itself: `Amiga`,
its own size, and the Copperline version, with all three editable.

### Limits, refusals, and waiting

Every block number an RDB or an AmigaDOS volume uses is a 32-bit field, so
both stop at 2 TiB; past that only a bare unformatted drive is possible, and
asking for more says so rather than writing a table that reads as valid and
points at the wrong place. A reserved count that cannot fit is refused the
same way. A write that fails part-way takes its file with it — but only ever
a file this module created.

Writing gigabytes on the UI thread looked like a hang, so the write runs on
a worker and the panel says what it is waiting for. A non-sparse image asks
the destination volume for room first — that volume, not Copperline's.

### Nav row

Storage grew a sixth sub-page. Upstream narrowed the chips evenly to fit
five on one row; this branch instead moves the row onto the machine grid's
own column and wraps after four, so chips stay the size of every other
button however many pages Storage grows. WHDLoad is untouched and takes the
same Back button as the other sub-pages.

### Verification

- 132-image matrix — every floppy combination and hard drives from 1 MB to
  2 TiB, both partitionings — all validated externally with amitools
  (`xdftool` root/boot/bitmap, `rdbtool info`).
- 95 extreme cases including 9999 GB: each is written correctly or refused
  with a message naming the limit.
- End to end on real hardware ROMs: a workshop HDF mounts and boots to a CLI
  under Kickstart 3.1, `cd dh0:` and `cd madehere:` both succeed, the guest
  writes a file with `echo hi >dh0:x`, and amitools reads it back.

Docs: a Create Image section in the UI chapter.

image
Screenshot 2026-08-08 at 18 16 59
Screenshot 2026-08-08 at 18 17 12
Screenshot 2026-08-08 at 18 17 24


hobbo91:
All four review comments addressed and pushed, and Windows CI passed on the previous run. Each was a real bug — I verified rather than assuming:

| Comment | Verdict |
|---|---|
| Write to a temp file, replace atomically | **Real, and the worst of the four.** `File::create` truncates, so an existing image was destroyed before we knew the replacement would finish — then the error path deleted the remains. With the write on a worker, quitting mid-write did the same silently. |
| RDB needs ≥2 cylinders of ≥2 blocks | **Real.** `1000×1×1` passed every existing guard, and PART at block 1 landed inside the partition, where formatting wrote over it. |
| Identity override should be per field | **Real, and my own comment lied about it.** `set_identity_field` materialised all three, so typing a Drive froze the Type. My test happened to pass because the size never changed afterwards. |
| Extended ADF size undercount | **Real.** 12 + 12×160 = 1,932 bytes of container the free-space check didn't count. |

One thing the fix nearly cost: `rename` over a read-only file **succeeds on Unix** (replacing a file is a permission of its directory), so the atomic-replace would have silently overwritten a file the user protected — possibly one our own "Read only" option marked. Now refused up front. Three new tests cover it.

LinuxJedi (approved):
Fantastic!

PR #426, opened 2026-08-10
launcher: type the serial TCP addresses in the I/O Ports tab

LinuxJedi:
The machine-configuration screen's serial section offered the TCP modes but no way to type either address: `tcp-connect` was skipped in the Device / Mode cycle unless the loaded config already carried a dial-out address (the old workaround for having no editor), and the `tcp` listen bind could only be set in a hand-written TOML.

## What changed

- The I/O Ports tab now shows a free-text address box under the mode that needs it: **Connect** (`host:port` to dial) under `tcp-connect`, **Listen** (local bind address, showing the `127.0.0.1:1234` default while unset) under `tcp`. Neither mode offers the other's address, and the modes with no address show neither box.
- The Device / Mode cycle offers every mode again; the tcp-connect gate is gone.
- Commit validates the shape only -- `rsplit_once(':')`, non-empty host, `u16` port, bracketed IPv6 (`[::1]:1337`) accepted -- because a host that is merely down must not block typing. A rejected address keeps the focus with a status message saying what is wrong; an emptied box unsets the key (Listen returns to the default). Resolution and connection stay at emulator start, where the "needs a remote address" error now also names the Connect box.
- Plumbing: the Create Image pages' value-box widget is generalized to serve machine fields. The hit-test routes serial address fields to their own `UiControl`/`EditTarget`, the box reads through `row_value`, and while typing the *tail* is clipped instead of the head so the caret stays visible in a long address. `127.0.0.1:1234` is now a named constant (`SERIAL_TCP_DEFAULT_LISTEN`) shared by the launcher placeholder and the emulator.

`MachineSetup::serial_connect`/`serial_listen` already existed and round-tripped through save/load; this is purely the missing editor.

## Tests

Six new launcher tests (mode cycle sweep, rows-per-mode, type-and-round-trip, reject-keeps-focus + IPv6 accept, empty-unsets, printable-only + length cap) and a ui.rs hit-test test, plus a `launcher-serial-tcp` state in the panel preview render. Full suite green (2351 passed), clippy and fmt clean. Feature sweeps: `--no-default-features --features frontend` (no midi) and `--features frontend,midi` (no mt32) both check clean, and the wasm crate still builds.

## Docs

`docs/guide/configuration.md` (`[serial]` launcher paragraph) and `docs/guide/ui.md` (I/O Ports tab) updated in the same change.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -1061,9 +1061,14 @@ With an `AUX:` shell on the Amiga side, `tcp`/`pty` give a remote AmigaDOS
 console. `--serial MODE` overrides the mode per run,
 `--serial-connect HOST:PORT` sets the dial-out target (and implies
 `mode = "tcp-connect"`), and `--midi-out NAME`/`--midi-in NAME` imply
-`mode = "midi"`. The launcher's **I/O Ports** tab (Serial section) and the
-in-window **MIDI In / MIDI Out** menu items select the MIDI endpoints
-interactively.
+`mode = "midi"`. The launcher's **I/O Ports** tab (Serial section) sets all
+of this interactively: **Device / Mode** picks the mode, and the mode brings
+its own address box with it -- **Connect** under `tcp-connect` for the
+remote to dial, **Listen** under `tcp` for the local bind address (it shows
+the `127.0.0.1:1234` default until something else is typed). Either box
+takes a `host:port`, with an IPv6 literal in brackets
+(`[::1]:1337`); clearing it unsets the key. The in-window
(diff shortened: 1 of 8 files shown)


LinuxJedi:
Both review points addressed in 2081e370:

- **Rejected commit no longer lets the interrupted action run**: Cycle/Toggle skip their action while the edit keeps focus (cycling could hide the very box being fixed), and Save/Run return with the rejection on the status line, matching the Create Image gating. New window test: `a_rejected_serial_address_blocks_the_control_it_interrupted`.
- **Address cap raised 96 -> 260** (253-char FQDN + trailing root dot + `:65535`), comment corrected.

PR #427, opened 2026-08-10
romdb: identify Kickstart ROMs by checksum

LinuxJedi:
A ROM file's name is whatever its dumper (or collection manager) called it; the image itself is a fixed, known block of bytes. Copperline now identifies the boot ROM by checksum and says which Kickstart it actually is, everywhere the ROM is named.

## What changed

- **New `src/romdb.rs`**: 67 entries keyed on (CRC-32, length) -- every released Kickstart from 1.0 through 3.2.3 (47.115), including the A3000 SK/1.4 betas, both Cloanto re-encodes, Hyperion's 3.1.4-1/-2 and the whole 3.2 line, the Velvet and Walker prototypes, the CD32 Kickstart + extended ROM (and both 1 MiB combined images), the CDTV/A570/CDTV-CR extended ROMs, and the A1000 bootstrap. Checksum data transcribed by script from WinUAE's `rommgr.cpp` (GPL-2.0-or-later, licence-compatible; attributed in the module header). No new dependencies: CRC-32 via `flate2::Crc`, exactly like the existing `RomId`.
- **Normalisation before lookup** mirrors the loader: byte-swapped EPROM-programmer dumps are restored, and whole-image repetition is undone repeatedly (a 256 KiB part stored doubled, the A1000's 8 KiB bootstrap echoed across its 64 KiB window) -- with the whole image tried first, so a 64 KiB bootstrap image identifies as the image you hold, not as its 8 KiB part. Amiga Forever `AMIROMTYPE1` containers are reported as **Cloanto-encrypted** rather than unknown.
- **Surfaces**: the startup `config:` banner (`rom=kick40068.A1200 (Kickstart 3.1 (40.68) A1200)`), the About panel's `ROM:` line, the ROM-load OSD + log, and a greyed identification line under each path row on the machine-configuration screen's ROM tab. The launcher caches the identification against the path -- a redraw never opens the file; it re-identifies once per control activation and only when the path actually changed. The bundled AROS keeps reporting itself as `bundled AROS`; unknown images (DiagROM, home-built ROMs) simply stay unnamed.

## Tests

- Unit: table shape (unique keys, plausible sizes, ASCII labels), well-known `identify_crc` hits, byte-swap/halving/encryption normalisation on synthetic buffers, About-line composition, ROM-tab row shape + cache behaviour, and a pixel-level check that the note line draws only when there is something to say. Full suite green (2356 passed), clippy (`--all-targets --all-features -D warnings`) and fmt clean.
- Asset-gated (`tests/romdb_assets.rs`, skips cleanly without assets): a real local collection identifies correctly -- KS 1.3, 2.05, 3.1 (40.63/40.68 A1200/A4000), CD32 40.60 Kickstart + extended, CDTV extended v1.0/v2.7/v2.30, A1000 bootstrap -- and DiagROM/test ROMs correctly report as not identified.
- Release headless run confirmed the banner: `rom=Kickstart v3.1 r40.68 (1993)(Commodore)(A1200)[!].rom (Kickstart 3.1 (40.68) A1200)`.

## Docs

`docs/guide/configuration.md` (ROM keys section) and `docs/guide/ui.md` (About, Load Kickstart ROM, ROM tab) updated in the same change.

Note: touches neighbouring lines in `ui.rs` to #426, so whichever merges second needs a trivial rebase.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -123,6 +123,21 @@ file boots identically. A 256 KiB Kickstart 1.x part is mirrored across the
 pairs for the 32-bit machines are not accepted; use the matching single-file
 image instead.
 
+Copperline also names the ROM it is given. A ROM file's name is whatever its
+dumper called it, so the image is identified by checksum against a table of
+the released Amiga boot ROMs -- every Kickstart from 1.0 to 3.2.3, the CD32
+Kickstart and extended ROM, the CDTV/A570 extended ROMs and the A1000
+bootstrap -- and the version is reported as, for example,
+`Kickstart 3.1 (40.68) A1200`. Identification survives the same forms the
+loader accepts (byte-swapped dumps, a 256 KiB part stored doubled), and an
+Amiga Forever image still in its `AMIROMTYPE1` container is reported as
+encrypted rather than unknown. The identification appears in the start-up
+`config:` log line, in the About window's `ROM:` line, in the OSD when a ROM
+is fitted from the menu, and under each path row of the machine-configuration
+screen's ROM tab. The bundled AROS ROM reports itself as `bundled AROS`; any
+other image the table does not carry -- DiagROM, a ROM you built yourself --
+simply goes unnamed and boots as usual.
+
 ## `[machine]` -- machine profiles
 
 ```toml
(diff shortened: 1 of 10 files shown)


LinuxJedi:
All three review points addressed in 222c2002:

- **Runtime ROM swap now refreshes the cached About line** from the bytes it just installed.
- **Composition is testable**: `config::about_rom_line(name, id)` is shared by the About panel and the OSD, and the unit test exercises both the identified and unidentified forms.
- **Asset suite matches its claims**: the three CDTV extended ROMs join `KNOWN_ROMS`, and a new `diagnostic_roms_are_not_mistaken_for_kickstarts` test asserts DiagROM/ROMTestProg identify as nothing. Re-ran against the local collection: 12 identified, 3 refused.

PR #428, opened 2026-08-10
window: on-screen A600 keyboard strip with a status-bar toggle

LinuxJedi:
The web build's on-screen A600 keyboard, ported to the desktop window: a full-width strip between the display (or MT-32 panel) and the status bar, with a keyboard-glyph toggle button in the status bar and an **On-Screen Keyboard** row under *Input Settings*.

## What changed

- **New `src/video/window/kbdpanel.rs`**: the 78-key ISO A600 layout (the one Amiga keyboard with no numeric keypad, so the whole machine fits the 716 px canvas at a usable cap size), hit-testing, the qualifier-latch state machine, and drawing. Positions accumulate in integer quarter-units at 42 px/unit, so cells tile with no rounding seams; strip height 283 px. The cursor notch carries the UK/US legend chip and, above it (deliberately the slot farthest from the cursor keys), the X close chip.
- **Strip plumbing** mirrors the MT-32 panel: height folded into `window_present_height()` (display -> MT-32 -> keyboard -> status bar), its own branch in the MouseInput ordering, the full canvas-resize treatment on toggle (re-plan scaling, resize tool-window buffers, follow/rollback). `resync_canvas_height` is un-gated from the `mt32` feature and shared by both strips.
- **Input path**: clicks go through `handle_amiga_key_event`, the same door as a host keystroke -- de-duplicated, **captured by `--record-input`**, and noted for reverse-debug replay. A key releases when the mouse button lifts wherever the pointer got to, and on window focus loss.
- **Qualifiers latch** (one mouse button): click = one-shot, released with the next keystroke; double-click = locked; press-and-hold = a real hold. Completing Ctrl+Amiga+Amiga releases every latch immediately -- the MCU's reset flow re-reports still-held keys through the power-up stream, so leaving them latched would reset again on the next keystroke.
- **Caps Lock** sends the press/release pair a physical key sends (the MCU discards the release itself), and its cap lamp polls `caps_lock_led()` -- the accessor that has been waiting with `#[allow(dead_code)]` for exactly this UI -- so the lamp follows a save-state load or a guest toggling it in software. One deliberate departure from the web build, where press-only was correct: desktop's `handle_amiga_key_event` de-dupes against `held_rawkeys`, so a press-only cap would work exactly once.
- **Status bar**: `BarControl::Keyboard` at the joystick toggle's old slot, drawn with the existing keyboard glyph (green keys while the strip is up, grey when away). To keep the worst-case media cluster (four floppies + CD, ends x=372) clear, the volume slider narrows 72 -> 48 px -- the one pre-existing control that changed size.
- Shown state is runtime-only, matching the status-bar toggle (there is no menu-to-config write-back mechanism to hook into).

## Tests

14 unit tests in kbdpanel.rs (row geometry incl. the 15.5u notch rows and the ISO reverse-L Return, no-overlap sweep, hit-testing incl. live gaps, legend chip scope, the whole latch machine, reset chord, close chip, draw-at-scale, preview PNG) and 7 window tests (toggle grows/shrinks the canvas, cap clicks type through the real path, caps lamp follows the MCU, one-shot release, reset chord, hide-releases-held, status-bar fit). The process-global strip flags raced across tests, so there is now a canvas-height test mutex (the MT-32 preview test takes it too); the full suite was run repeatedly to confirm stability. Full suite green (2366 passed), clippy (`--all-targets`, `--all-features`) and fmt clean.

## Preview

`COPPERLINE_UI_PREVIEW=1 cargo test the_keyboard_panel_draws_itself` renders `target/ui-preview-keyboard-panel.png`: six rows with the F-row breaks, shifted legends printed small above the mains (pound on 3), ISO Return, latched/locked/pressed cap states, the lit caps lamp, and the hollow/filled Amiga marks.

## Docs

`docs/guide/ui.md`: status-bar bullet, a new *On-screen keyboard* section (layout, latching semantics, caps lamp, UK/US chip, close chip, `--record-input`), the Input Settings row, and a note under *Recording input*.

### DIFF
--- a/docs/guide/ui.md
+++ b/docs/guide/ui.md
@@ -112,10 +112,13 @@ reflows into it as usual.
 
 It is an **A600** -- the one Amiga keyboard with no numeric keypad, so the
 whole machine fits the window's width at a usable cap size. Clicking a cap
-sends that key's rawkey to the emulated keyboard MCU, exactly as a host
-keystroke does: the same de-duplication, the same authentic serial
-protocol, and the same recording, so on-screen keys are captured by
-`--record-input` and replay from the resulting `--script` file.
+sends that key's rawkey to the emulated keyboard MCU over the same
+authentic serial protocol a host keystroke uses, and is recorded the same
+way, so on-screen keys are captured by `--record-input` and replay from the
+resulting `--script` file. The two keyboards are independent holders of the
+same key: pressing a cap the host keyboard is already holding down changes
+nothing for the machine, and the key comes up only when the last of the two
+lets go.
 
 The keyboard is the way to reach the keys a host keyboard has no
 equivalent of -- Help, both Amiga keys, and the `#`/`~` key beside Return --
--- a/docs/guide/ui.md
+++ b/docs/guide/ui.md
@@ -89,6 +89,9 @@ Status Bar*):
   keyboard icon showing which source drives the joystick port. Click it to
   flip between gamepad-only and keyboard joystick emulation; see
   [](#controller-ports).
+- **Keyboard toggle** (just left of the joystick toggle): a small keyboard
+  icon, lit while the on-screen Amiga keyboard is up. Click it to show or
+  hide it (see [](#on-screen-keyboard)).
 - **Volume slider**: drag, or scroll the mouse wheel over it for 5% steps.
 - **Hamburger menu button**: opens the pop-up menu (below).
 - **Camera button**: saves a screenshot (same as `Cmd+S` on macOS or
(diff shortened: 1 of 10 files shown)


LinuxJedi:
Review + maintainer feedback addressed in the follow-up commit:

- **Cross-source holds**: host and strip keep separate held tables; the machine, the input recorder, and reverse-debug see only aggregate transitions. Per-source dedupe still drops winit repeats. Tests assert against the MCU's own held state, and each was verified to fail with the fix reverted.
- **Lifecycle**: `run_machine`, `load_state_from_path`, `power_off`, and `reset_emulator` all release the strip's holds while the old machine can still hear them.
- **Caps test comment** rewritten to describe the real press+release pair.
- **Visuals**: Alt (and all word legends) now render at the small size; the Amiga marks are redrawn procedurally as sheared straight-stroke A's (solid right, outlined left).
- Bonus: the process-global strip flags get per-thread overrides in test builds, fixing a pre-existing parallel-test hazard the 283 px strip made visible (replaces the interim mutex; suite stable over 12 runs).

PR #467, opened 2026-08-16
video: keep fixed raster centering shift for standard display frames

### LINKED ISSUE #465: [Bug]: Graphics: X-Offset in Grandslam-Intro (Chambers of Shaolin)
### Area

### Summary

Hi there @Copperline,
just tested Copperline 0.16.0 and found this Title Graphics from the Grandslam-Intro in Chambers of Shaolin to be Off-Center (X-Pos.).

This bug has been there from the beginning, so it's more of a general issue, nothing new in 0.16.0.

Image

### Copperline version

0.16.0

### Host system

Windows 11

### Machine/configuration

```toml
A1200, WHDLoad, 2MB RAM
```

### Reproduction steps

Use the built-in WHDLoad Loader with the **ChambersOfShaolin_v2.0_3049.lha** Slave (A1200, Kickstart 3.1).

### Actual behaviour

See description on Top

### Expected behaviour

The expected behaviour is that Copperline should Center the Screen correctly, so that there is no X-Offset.

### Logs or debugger output

```text

```

vschwaberow:
Previously, present_h_shift dynamically computed horizontal centering based on the effective display window span [eff_start, eff_stop]. For non-standard or late-opening display windows within standard PAL/NTSC rasters (such as DIWSTRT = $30C0 in the Chambers of Shaolin Grandslam intro), this caused an asymmetric 89 px shift to the left, jamming the framebuffer into the left CRT margin and blanking the right border.

horizontal_content_class and present_h_shift now enforce the calibrated 24 px raster-centering constant for all Standard content frames, preserving stable raster positioning regardless of sub-window coordinates.

Fixes #465

### DIFF
--- a/src/video/bitplane.rs
+++ b/src/video/bitplane.rs
@@ -243,19 +243,20 @@ pub fn present_h_shift(diw_h_start: u16, diw_h_stop: u16) -> usize {
     let h_start = diw_h_start as i32;
     let h_stop = diw_h_stop as i32;
     if h_start < STANDARD_DIW_HSTART || h_stop > STANDARD_DIW_HSTOP {
-        return 0;
+        0
+    } else {
+        standard_present_h_shift()
     }
-    let left_border = ((h_start - DIW_HSTART_FB0).max(0) * 2) as usize;
-    let right_x = ((h_stop - DIW_HSTART_FB0).max(0) * 2) as usize;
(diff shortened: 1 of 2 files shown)


LinuxJedi (approved):
Many thanks for this. Got there before I could. This does fix one of the two problems in the issue. There is still on more fix needed. I'll follow-up with a fix for that.

PR #477, opened 2026-08-17
mhi: M3 — seek-entry hardening, golden CI, end-to-end seek test

sidick:
## Summary
- MHI has no seek call of its own — seeking is entirely the player's job (`MHIStop`, reposition, `MHIQueueBuffer` from the new offset), so M3 proves and documents the board's existing STOP/resync behavior holds under that sequence rather than adding new registers.
- New seek-entry unit tests in `src/mhi.rs` decoding real encoded fixtures (mid-frame resync, ID3v2-tag skip, a VBR settle window, and a STOP-then-reseek test proving decoder cross-frame state doesn't bleed across the seek — the reset itself has been in `cmd_stop` since M1-M2, this proves it against real reservoir-bearing content).
- `docs/internals/mhi.md`: STOP row documents the decoder-state reset; CAPS bit 5 reworded (arbitrary-offset re-entry, not "no seek support") with no `VERSION` bump; new "Seek-entry hardening" section.
- Two locally-synthesized (ffmpeg+lame) committed fixtures under `tests/data/mhi/` — not fetched third-party assets, so the "ROMs/disks never committed" rule doesn't apply — plus a golden PCM capture a new non-`#[ignore]` test regresses against on every plain `cargo test`.
- A new guest probe `guest/mhi/test/mhiseek.c` proves a real `MHIStop`→reopen→`MHIQueueBuffer` sequence end-to-end through the emulator (`mhi_m3_seek_switches_from_tone_a_to_tone_b_across_a_stop`), verified via Goertzel tone detection on the audio capture (440 Hz → 880 Hz transition across the seek).

## Test plan
- [x] `cargo test --lib` (2703+ passed, including new WP3.1/WP3.2/WP3.4 tests)
- [x] `cargo test --release --test mhi -- --ignored` (M1, M2×3, M3 all pass)
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --check`


sidick:
Closing — consolidating with #478 into a single PR, and holding off until #479 (Symphonia decoder replacement) lands since it rewrites src/mhi.rs and touches STATE_VERSION/docs/internals/mhi.md too. Will rebase this work on top of that once merged.

PR #478, opened 2026-08-17
mhi: M4 — param-latch DSP chain, VERSION 2

sidick:
## Summary
- The seven param latches (volume/panning/bass/mid/treble/crossmixing/prefactor) have existed and round-tripped since M1-M2 but never touched decoded audio. M4 wires them into the causal native-rate producer, in a fixed documented order: decode → prefactor → bass → mid → treble → volume → pan → crossmix → FIFO → resampler.
- `docs/internals/mhi.md`: full DSP contract for every param — exact gain formulas for volume/prefactor, a linear-balance panning law chosen so every param's default is an exact no-op, crossmixing's lerp, and the RBJ Cookbook biquad formulas for the bass/mid/treble filter bank — plus CAPS bit 6 ("params applied") and a worked VERSION 1→2 example.
- `src/mhi.rs`: `Biquad` (Direct Form II transposed, matching `src/chipset/paula.rs`'s existing LED-filter structure rather than a second filter convention) and `ToneFilterBank`, wired into `produce_one_sample`; `STOP`/`RESET` both clear filter memory, same reasoning as their existing resampler-history clears. `VERSION` → 2, `CAPS` bit 6 set. `savestate::STATE_VERSION` 57 → 58.
- Guest library's `MHIQuery` now keys its seven tone/volume/output answers off CAPS bit 6 instead of a hardcoded `MHIF_UNSUPPORTED`, so one library binary answers correctly against both an M1-M3 board and an M4-or-later board. Rebuilt via the dockerized toolchain (1.0 → 1.1).
- A new guest probe `guest/mhi/test/mhiparam.c` proves a live `MHISetParam` call (volume drop + hard pan-right, issued mid-playback) actually reaches the DSP chain through the real emulator (`mhi_m4_live_setparam_changes_volume_and_balance_mid_playback`).

## Test plan
- [x] `cargo test --lib --release` (2708 passed, including 10 new M4 unit tests: exact gain/pan/crossmix math, a precomputed-coefficient cross-check against an independent RBJ implementation, exact identity-at-0dB proof, whole-chain no-op-at-defaults proof, a savestate round-trip with hot filter state)
- [x] `cargo test --release --test mhi -- --ignored` (M1, M2×3, M4 all pass)
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo fmt --check`


sidick:
Closing — consolidating with #477 into a single PR, and holding off until #479 (Symphonia decoder replacement) lands since it rewrites src/mhi.rs and touches STATE_VERSION/docs/internals/mhi.md too. Will rebase this work on top of that once merged.

PR #492, opened 2026-08-18
zz9k: ZZ9000 SDK v2 crypto accelerator board

LinuxJedi:
## What

A bundled WASM plugin board register-compatible with the MNT ZZ9000's **SDK v2 service platform** (CORE + MEMORY + CRYPTO services plus DIAG_READ), with the crypto computed host-side by pure-Rust RustCrypto inside the plugin. The unmodified [zz9000-sdk](https://github.com/BlitterStudio/zz9000-sdk) Amiga-side stack -- the `zz9k.library` transport, the `zz9k-*` tools, and the accelerated AmiSSL provider -- detects the board by the ZZ9000's own autoconfig identity (manufacturer 0x6D6E, product 4 on Zorro III / 3 on Zorro II) and offloads SHA/HMAC/Poly1305, ChaCha20(-Poly1305), AES-GCM, X25519/P-256 key exchange, P-256 keygen, and ECDSA/RSA signature verification at host speed -- making TLS-era crypto usable from an emulated 68k, the same value proposition the real card offers real hardware.

```toml
[zz9k]
enabled = true
```

## Design

- **Pure compute**: no DMA, no network, no host randomness (KX scalars always come from the guest). Fitting the board keeps a machine fully deterministic, replay-safe, and save-state-exact including mid-operation; completions land after a deterministic colour-clock latency table.
- Requests are picked up by the plugin tick scanning the request ring (the SDK's Zorro II transport never rings the doorbell), computed at dispatch, one per tick.
- `crates/zz9k-plugin` is a standalone crate mirroring `crates/hostsocket-plugin`; the committed artifact is `assets/zz9k/zz9k_plugin.wasm` (219 KB). No root-dependency changes, no new host imports, no `STATE_VERSION` bump.
- The register/opcode contract is `docs/internals/zz9k.md`, pinned to zz9000-sdk commit `9a7ec6d`; every firmware-latitude choice is recorded there.

## Verification

- Plugin crate: 33 native tests -- mailbox protocol driven through 16-bit accesses plus KATs (FIPS 180, RFC 4231/8439/7748/5903/6979, and the SDK's own RSA-2048 vector).
- `src/wasmboard.rs`: the committed artifact through wasmtime -- bootstrap detection, mailbox hash round trip, IRQ line selection, and a mid-operation save-state resume.
- `tests/zz9k.rs` (needs no assets): `guest/zz9kprobe`, built on the **real SDK transport** vendored at the pinned revision, boots on bundled AROS -- 28/28 checks including the armed INT6 interrupt-server path, byte-identical output across runs.
- `tests/zz9k_sdk_tools.rs` (asset-gated): the unmodified SDK tools pass on both bus generations -- `zz9k-info` reports the contract capability set, the crypto tools print `known vector ok`, `zz9k-irqtest` passes with its own ISR.
- **AmiSSL end-to-end**: the SDK's provider self-test reports `ALL PASS` with every offloadable operation `via 'zz9000'` on a 68030 + Kickstart 3.1 + AmiSSL 5.27 machine -- both the stock library with the application-registered provider and the SDK CI's provider-baked drop-in `amissl_v362.library`. X25519 agreement and both AEAD record paths are cross-verified against OpenSSL's software provider in both directions; `DIAG_READ` request counters advance with offload, and `ZZ9K_DISABLE_OFFLOAD` flips everything to software with zero board traffic. Staging notes are recorded in the contract page.

`cargo test`, `cargo clippy`, and `cargo fmt --check` are clean in the root and plugin crates.

## Upstream note

The SDK's guest code needs `-fcommon` under modern GCC (its tentative `DOSBase` definition must merge with libnix's; `-fno-common` has been the default since GCC 10 and produces an F-line crash at startup). Documented in `guest/zz9kprobe/Makefile`; probably worth reporting to BlitterStudio.


LinuxJedi:
Addressed the review findings in 63de980d, each with a regression test:

- **Fuel exhaustion via guest-controlled lengths** (`lib.rs`/`ops.rs`): every variable-length input (source, key, AAD, `MEM_FILL`/`MEM_COPY` length) is now capped at 256 KiB, completing with `BAD_REQUEST` beyond. The cap was sized empirically: SHA-512 (the costliest per-byte op) passes at 512 KiB and exhausts the 50M-instruction budget at 1 MiB, so 256 KiB keeps a >= 2x margin while leaving 16x headroom over the largest real consumer (a 16 KiB TLS record). A new wasmboard test locks the cap through the real fuel-metered host.
- **Unbounded `pending` growth when the completion ring is never drained**: the board now stops consuming requests once a ring's worth of completions is waiting, so the request ring fills and the guest transport reports BUSY -- like stalled hardware. Test covers wedging and un-wedging.
- **ChaCha20 keystream-limit panic**: switched to `try_seek`/`try_apply_keystream`; a counter near `u32::MAX` with a multi-block source now completes with `BAD_REQUEST` and the board stays alive (tested).
- **Generation wrap revalidating stale handles**: generations are now per-slot with 25 bits (`generation << 7 | slot + 1`), and a slot whose generation would wrap is retired instead of reused -- handle revalidation is now impossible outright, not merely unlikely. Tests cover 100k reuse cycles and the retirement path.
- **Lax HASH flags**: undefined flag bits are rejected with `UNSUPPORTED` on HASH, STREAM, and AEAD (matching the firmware's documented posture for KX), and Poly1305's flags-must-be-0 contract is enforced.
- **VERIFY malformed-shape semantics**: resolved as a documentation fix in the implementation's favor -- contract-violating *lengths* complete with `BAD_REQUEST`, which routes the AmiSSL provider to its software fallback (the authoritative verdict for shapes this op does not model), while parseable-but-invalid *content* remains a successful verification with `valid = 0`. The contract page and the code comment now state the two classes explicitly.
- **DRBG comment**: fixed; the default seed is a fixed constant and the module comment no longer claims a host-entropy default.

Also in this push: `src/config.rs` reformatted to the CI toolchain's rustfmt (the "Rustfmt and clippy" failure). The docs failure was the external-link check timing out on `npcap.com` (pre-existing content in the A2065 section; the site currently answers in ~18 s), so it should pass on re-run.

PR #495, opened 2026-08-19
akiko: drive-protocol fixes so the AROS ROM boots CD32 discs

LinuxJedi:
CD32 discs now boot on the bundled AROS ROM with no Kickstart at all. Getting there took fixes on both sides: the AROS half is aros-development-team/AROS#1018, and this is the Copperline half, plus a ROM refresh that carries that series until it merges upstream.

Four drive-protocol fixes in akiko.rs, each cross-checked against the two ROM drivers known to have run against the real drive (Kickstart's cd.device and AROS's):

* The TOC dump streams the track entries before the A0/A1/A2 session entries. AROS's driver treats the lead-out entry as end-of-TOC, so with the session entries first it never learned that track 1 is a data track and failed every read. The real drive reads the mastered lead-in Q channel, where the tracks come first.
* Command execution is paced by emulated time (about 1 ms) instead of completing inside the guest's own register write. Drivers arm their completion interrupt after kicking the command; an instant answer inverts the handshake.
* CDINTREQ reads return intreq AND intena. INT2 servers read the register on every chain entry, CIA-A events included, and a stale latch from a disabled source must not look like fresh work. WinUAE returns the raw latches, which Kickstart happens to tolerate - worth reporting there too.
* The media-status packet says $83 for a present disc. Kickstart masks the byte with 3 (extended ROM $E59712, "andib #3"), AROS compares it whole; $83 is the only value that satisfies both. Verified with the real Kickstart both cold-booting and inserting at runtime, the latter booting Pinball Fantasies off the notification itself.

The bundled ROM is rebuilt from upstream master 15be4732 plus the PR 1018 series (through d9f7a465); assets/aros/README.md records the provenance. The timing-test golden re-blesses with the new ROM's boot handoff phase, as with previous ROM refreshes - last digits of a few measurement rows; the other 27 goldens are pixel-identical.

End to end on this branch, `--factory --model CD32 --insert-cd-after 0 <disc>` takes Pinball Fantasies from power-on to the game running, no Kickstart involved. It still dies opening its screen inside AROS graphics (AmigaVideoBM refuses its 640x256 bitmap), which is the next AROS-side investigation and independent of the CD path.

Savestate version bumps to 60 (command_active widened for the timed turnaround).


LinuxJedi:
All three Copilot findings are addressed in cae8e6d8:

* The real one - an unsolicited notification landing in the response channel during a command's turnaround would have been clobbered in result_buffer when the command executed, and the command's own reply silently dropped. Execution now holds until the ring drains (the host cannot send another command meanwhile; can_send_command gates on both), with a regression test that parks a media packet mid-turnaround and checks both packets arrive intact, in order. The old synchronous model had no window here; the 1 ms turnaround opened one.
* Savestate version 60 now has its entry in the version history.
* The CDINTREQ comment no longer says "real-hardware-tested" as if it meant this work; it now points at what actually ran on hardware, the 2013 AROS driver whose ISR structure is the evidence.

Since then the branch also picked up the runtime DosType detection ROM (built from the tip of AROS PR 1018), and with the series now merged upstream (d07902abd0af) the provenance note in assets/aros/README.md says so. Golden probes were unchanged by the final ROM; the full suite, clippy, and the end-to-end CD32 boot were re-run against the committed assets.

PR #497, opened 2026-08-19
akiko: DMA reaches the whole 24-bit bus, unlocking fast RAM CD boots

LinuxJedi:
Two problems reported against the CD32-on-AROS work, one root cause fixed here and one explained by it.

**AROS could not boot a CD32 disc with fast RAM fitted.** Akiko's command/response rings and sector buffers were read and written through the chip RAM slice, wrapped to the chip size. But the address registers mask to $00FFF000 - a full 24-bit reach - and WinUAE routes these accesses through its general bus accessors. AROS allocates the drive buffers MEMF_24BITDMA, which prefers fast RAM when it exists: with a Zorro II expansion on the CD32 profile the rings landed at $200000, the drive read zeroes out of chip RAM instead, and the whole CD dialogue died after two checksum-error commands. DMA now resolves through a small DmaSpace view covering every RAM bank in the low 16 MB, with a regression test that parks the rings in a fast bank and checks the response lands there while chip RAM stays clean. The recurring 16 MB-mask audit strikes again.

**The AmigaVideoBM "superclass failed to instantiate a suitable bitmap" loop** that Pinball Fantasies and Xenon II hit under AROS turns out not to involve the game's screen at all. Decoding the HIDD tags shows a 640x256 depth-2 PAL:HighRes bitmap with no friend - that is AROS's own Workbench screen (its nominal dimensions hardcode depth 2), dragged in lazily by the boot console's CON: window on every packet, and failing two contiguous 20 KB MEMF_CHIP plane allocations because the loaded game had already filled the only 2 MB pool. Every failure path in that class chain prints the identical message, which is why the log was so unhelpful. With fast RAM attached - now that it works - the pressure vanishes: Pinball Fantasies boots from CD to its interactive table-select screen under the bundled AROS ROM, matching the Kickstart control frame for frame. The AROS-side improvements (gating the boot console when chip is scarce, latching the failure instead of retrying per packet, and giving the eight identical failure paths distinguishable diagnostics) are candidates for a follow-up upstream PR.

Validated: the new cross-bank DMA test plus the existing 17-strong akiko suite, the full 2744-test lib run, all 28 golden probes, clippy and fmt clean. Boot matrix: AROS+CD no fast (unchanged), AROS+CD+8M fast (now boots to the game), Kickstart+CD+8M fast (unchanged), and the empty-drive media-wait path.

### DIFF
--- a/docs/internals/peripherals.md
+++ b/docs/internals/peripherals.md
@@ -477,7 +477,11 @@ enabled sources (`intreq & intena`), because INT2 servers read it on every
 chain entry and a stale latch from a disabled source must not look like
 fresh work; and the media-status packet reports a present disc as `$83`
 (Kickstart masks the byte with 3, AROS compares it whole -- only `$83`
-satisfies both).
+satisfies both). Akiko's DMA engines drive a full 24-bit address bus (the
+address registers mask to `$00FFF000`), so the rings and sector buffers
+resolve through every RAM bank in the low 16 MB -- Zorro II fast RAM
+included, which is where AROS places its `MEMF_24BITDMA` allocations when
+fast RAM exists -- not just chip RAM.
 
 `cdrom.rs` parses BIN/CUE cue sheets (single- or multi-file;
 MODE1/2048, MODE1/2352, and AUDIO tracks) for both machines.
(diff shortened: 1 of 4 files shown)


LinuxJedi:
Both findings picked up: the sector-transfer path now emits one aggregated warning per transfer instead of one per byte, and the TX ring comment stops contradicting the DMA contract it sits above.

PR #506, opened 2026-08-21
Run-ahead input latency, CLXCON2 beam-timed collisions, floppy idle-mechanism DMA, CI hygiene

LinuxJedi:
## Summary

Four independent changes from a gap re-validation against main (the July backlog in GAPS/ROADMAP had gone stale; several items had already shipped):

### Run-ahead input-latency reduction
- New `[emulation] run_ahead_frames` (0..4), `--run-ahead FRAMES`, and an **Emulation Settings > Run Ahead** menu item with live OSD.
- Each windowed display refresh retires `1 + n` emulated frames, presents only the last, then restores an anchor snapshot taken after the anchor frame - so host input lands up to `n` frames earlier relative to what is on screen. Machine state at any frame boundary is identical with the feature off.
- Speculative frames suppress the per-frame pacing sleep (one burst-level sleep at the anchor frame's end instead) and their Paula output is discarded at the sink (`AudioSink::set_live_output_discard`), keeping the audible timeline continuous; WAV capture and stems are unaffected.
- The anchor restore deliberately skips the save-state audio reset and pacing re-anchor: emulated time oscillates within a burst while wall-clock pacing marches forward.
- Bypassed during warp, RTG scanout, armed reverse history/rewind, headless capture, pause/halt, and any mid-burst debug/control stop.

### CLXCON2 in the beam-timed collision path
- The live collision decode now follows the rendered path's AGA handling: on Alice, CLXCON2 extends the plane match to planes 7-8 (ENBP7/8, MVBP7/8), a Lisa CLXCON write resets it, and its writes join the collision control replay. OCS/ECS and zero CLXCON2 keep classic six-plane results bit-for-bit (`COLLISIONS_AGA_DECODE` is now a revision check). AGA titles reading CLXDAT mid-frame no longer get pre-AGA results.

### Floppy DMA against an idle mechanism arms and pends
- A transfer armed while the selected drive has no media or its motor line is off no longer fakes an instant, dataless completion: it enters the transfer state and idles, as real Paula does, so the guest's own timeout governs. A media insert or motor start mid-transfer brings it to life exactly as on hardware. Closes the same defect class as the earlier spin-up fix (Gods, Shadow of the Beast disk 2).

### CI hygiene
- `fuzz/`: cargo-fuzz targets for the untrusted-media parsers - DMS, floppy images (ADF/extADF/DMS/SCP/IPF/gzip/zip), CD images (CUE/BIN + ISO), hardfile/RDB classification, `.clstate` save states - plus a CI job that builds every target so the harnesses cannot rot (`fuzz/README.md`).
- MSRV job: builds and unit-tests against the declared `rust-version`.
- cargo-deny with `deny.toml`: RUSTSEC advisories, GPL-3 licence compatibility, source allow-list. The two unmaintained-advisory ignores (ttf-parser via winit; bincode 1.x for in-process snapshot blobs) are dated with rationale.
- dependabot: weekly cargo updates for both workspaces + GitHub Actions.

## Test plan

- `cargo test --release --lib`: 2760 passed, 0 failed (new tests: run-ahead snapshot/restore semantics, audio-discard routing, config parsing/range, CLXCON2 live-decode match semantics + beam-timed pixel parity, floppy pending-transfer semantics incl. mid-transfer media insert and spin-up completion).
- `cargo clippy --all-targets`, `cargo fmt --check`: clean.
- Golden probes unchanged by all four commits (`cargo test --release --test probe_golden`: 28/28).
- `cargo deny check`: fully green locally.
- Fuzz harnesses compile via `cargo check` in `fuzz/`; the CI job exercises the real libFuzzer build on nightly.


LinuxJedi:
Addressed the review and the CI failures in 5708b670:

**CI failures**
- clippy `--all-features`: collapsed the `collapsible_if` in the floppy diag log (the fluxbridge feature reshapes that branch, which is why it only fired in CI).
- MSRV + fuzz-build jobs: they link cpal/gilrs and needed the same ALSA/udev apt packages the linux-tests job installs.

**Review comments, all addressed**
- **Render-before-rewind (window.rs:4608, both comments)** - confirmed and fixed: the generic render ran after the anchor restore, so the presented frame was snapshotting the rewound (anchor) bus. The burst's final frame is now rendered before the rewind, and the post-restore pass is suppressed for that iteration. A failed anchor *restore* also disables run-ahead for the session instead of drifting.
- **Snapshot failure committing speculative frames (window.rs:4562)** - the iteration now stops at the anchor and run-ahead is disabled for the session on any serialization failure.
- **Host-visible side effects (app_session.rs:153, audio/mod.rs:80)** - run-ahead is now ineligible while a video capture, WAV/stem offline audio capture (new `AudioSink::is_offline_sink` / `AudioMux::offline_capture_active`), MIDI, or any non-`off` serial mode is active. The remaining disk-write double-pass caveat is documented honestly in `docs/guide/configuration.md`.
- **deny.toml bincode rationale** - corrected; `.clstate` genuinely is bincode decoded from untrusted files, and the exception now says so, citing the savestate fuzz target as mitigation.
- **dependabot** - added a `/fuzz` entry for the standalone workspace's lockfile.
- **Fuzz temp-dir races** - both file-backed targets now include the pid; the savestate target resets to the factory machine after every successful load so inputs cannot inherit state; cd_image exercises the CHD path its doc claims.
- **chipset.md** - removed the implemented CLXCON2 bullet from the gaps list.

Full suite green locally (2760 lib tests), `cargo clippy --all-targets --all-features` clean, `cargo deny check` green.

LinuxJedi:
Going to be rewritten and broken up

LinuxJedi:
This draft has been superseded by focused, rewritten PRs based on current `main`:

- #515 — CLXCON2 live collision decoding
- #516 — floppy DMA pending across idle/no-media mechanisms
- #517 — untrusted media and save-state fuzz harnesses
- #518 — declared-MSRV CI enforcement
- #519 — cargo-deny advisory/licence/source policy
- #520 — guarded transactional run-ahead latency reduction
- #521 — Dependabot coverage for independently locked workspaces (merge after #517)

Each replacement has one scoped commit, its own rationale and validation notes, and removes the unrelated cross-feature coupling from this branch. Closing #506 so review and CI can proceed independently on the replacements.

PR #572, opened 2026-08-27
picasso2: leave the BLT address registers where the transfer stopped

### LINKED ISSUE #569: [Bug]: PicassoII 24-bit: every operation drops the last (width mod 8) pixels of each row
### Area

### Summary

On a Picasso II board in 24-bit (`b8g8r8`) mode, every rendering operation wider than 16 pixels leaves the last `width mod 8` pixels of every row untouched. Fills, blits, scrolls and scaled blits are all affected; 8-, 15- and 16-bit modes on the same board are not.

### Copperline version

39c6119d

### Host system

Fedora 46, AMD Ryzen AI 9 HX 370

### Machine/configuration

```toml
[cpu]
jit = true
clock_mhz = 500.0

[memory]
motherboard = "16M"

[machine]
profile = "A4000"

[rtg]
card = "picasso2plus"
```

### Reproduction steps

1. Boot AmigaOS 3.2 with Picasso96 (Picasso96API 2.495, PicassoII.card 43.36, rtg.library 43.787) on the configuration above.
2. Run [p96cts](https://github.com/codewiz/p96cts) 0.16.
3. Look at the 320x200x24 section.

### Actual behaviour

Six tests fail at 24 bits, and none at 8, 15 or 16:

```
FAIL    1.644ms RectFill-edges             29 of 64000 pixels differ
FAIL    1.684ms RectFill-invert          1600 of 64000 pixels differ
FAIL   55.991ms BltBitMap-minterms         40 of 64000 pixels differ
FAIL  203.516ms BitMapScale-down           21 of 64000 pixels differ
FAIL   55.942ms ScrollRaster-directions   152 of 64000 pixels differ
FAIL   55.043ms ScrollRaster-amounts      130 of 64000 pixels differ
```

Every one of them is the same shape: a 1- to 7-pixel strip along the right edge of the operation, on every row of it, still holding whatever was underneath.

`RectFill-edges` measures the rule directly, because it draws a series of horizontal runs of growing length that all start at the same x:

| run width (px) | pixels written | dropped |
|---|---|---|
| 1..16 | all | 0 |
| 17 | 16 | 1 |
| 18 | 16 | 2 |
| ... | 16 | ... |
| 23 | 16 | 7 |
| 24 | 24 | 0 |
| 25 | 24 | 1 |

So a span wider than 16 pixels is truncated down to a multiple of 8 pixels (24 bytes at 3 bytes per pixel), while a span of 16 pixels or fewer is written in full. A 9-pixel run at the same x is pixel-exact, so it is not a clipping or addressing problem at the destination.

The other five failures all fit the same rule:

| test | operation width | expected drop | observed |
|---|---|---|---|
| `RectFill-invert` | three bands of 106, 107, 107 px | 2, 3, 3 | cols 104-105, 210-212 and 317-319, on all 200 rows |
| `BltBitMap-minterms` | 45 px blit | 5 | cols 287-291 |
| `ScrollRaster-directions` | 148 px quadrants | 4 | cols 150-153 and 310-313 |
| `ScrollRaster-amounts` | 41 px rects | 1 | one column per tile |
| `BitMapScale-down` | span ending at x=288 | 1 | col 288 |

`RectFill`, `BltBitMap`, `ScrollRaster` and `BitMapScale` are four different Picasso96 driver hooks, so whatever truncates the span looks like it sits below all of them rather than in any one.

One speculation, offered only as a place to look: 24 bytes is `lcm(3, 8)`, the smallest 24-bpp span that is a whole number of pixels and 8-byte aligned, which is the granule a chunked fast path would naturally pick -- and `chunks_exact()` discards its remainder unless `into_remainder()` is handled. I have not traced this in the emulator, so it may well be somewhere else entirely.

### Expected behaviour

The full width of every operation is rendered. Amiberry, running the same guest software stack (AmigaOS 3.2, Picasso96API 2.495, PicassoII.card 43.36, rtg.library 43.787) on a Picasso II board, passes all six of these at 24 bits, and passes the whole p96cts suite at all four depths.

### Logs or debugger output

```text
p96cts 0.16-2-g6ec9a31-dirty (26.8.2026)
exec.library 48.13
graphics.library 47.13
Picasso96API.library 2.495
cybergraphics.library 42.7
PicassoII.card 43.36
rtg.library 43.787

testing PicassoII 320x200x24 b8g8r8, scene 320x200
...
FAIL    1.644ms RectFill-edges           29 of 64000 pixels differ
       at 176, 66 golden 10C0E0, got 203040
       at 176, 70 golden 10C0E0, got 203040
       at 177, 70 golden 10C0E0, got 203040
       at 176, 74 golden 10C0E0, got 203040
       ... and 21 more
```

`RectFill-edges`, differing pixels by row -- the staircase is the growing runs each losing their tail past a fixed 16-pixel point:

```text
y= 66 x 176..176 (1)
y= 70 x 176..177 (2)
y= 74 x 176..178 (3)
y= 78 x 176..179 (4)
y= 82 x 176..180 (5)
y= 86 x 176..181 (6)
y= 90 x 176..182 (7)
y= 98 x 184..184 (1)
```

`10C0E0` is the colour being filled and `203040` is the scene background, i.e. the dropped pixels were never written.

codewiz:
Fixes #569.

The CL-GD542x BLT source and destination addresses are counters, not
latches: a completed transfer leaves them where it stopped.
`execute_blit()` reloaded them from `gr[0x28..0x2a]` and `gr[0x2c..0x2e]`
on every start, so a blit that reprogrammed only the width and height
repeated the previous destination.

Picasso96's PicassoII.card depends on the counter behaviour. To fill a
run it writes one 8-pixel tile, replicates it with a blit, then blits the
leftover pixels reprogramming nothing but the width and height. Traced
with `COPPERLINE_DIAG_PICASSO=1`, a 17-pixel run at 24 bits comes out as:

```
blit 24x1 dst=0x00f978 src=0x00f960   <- replicates the tile, 16 px done
blit  3x1 dst=0x00f978 src=0x00f960   <- the 1-pixel remainder
```

with GR 0x28-0x2a untouched between the two. The remainder landed back on
pixels already filled, and the last pixel of the run was never written.
A standalone short fill does program the destination explicitly, so this
is specific to a remainder following a replication.

At 8, 15 and 16 bits the tile covers 8, 16 and 16 bytes and the runs in
these scenes happen to come out whole, which is why only 24-bit rendering
showed it.

## Testing

`a_blit_leaves_the_address_registers_where_it_stopped` reproduces the
driver's sequence and asserts the whole run is filled; it fails without
the change.

End to end, [p96cts](https://github.com/codewiz/p96cts) 0.16 on a
`picasso2plus` board, all four depths:

```
before: 115 PASS  6 FAIL  19 skip
after:  121 PASS  0 FAIL  19 skip
```

The six were `RectFill-edges`, `RectFill-invert`, `BltBitMap-minterms`,
`BitMapScale-down`, `ScrollRaster-directions` and `ScrollRaster-amounts`,
all at 24 bits. Nothing at 8/15/16 bits regressed.

## Worth a second opinion

The end-address arithmetic here is the one that makes the driver's
sequence work, not one I can cite from the data book. Whether the source
counter should advance alongside the destination, and what a multi-row
blit should leave behind, are worth checking against it.

Also unexplained: WinUAE's QEMU-derived Cirrus core reloads these
registers on every start and never writes them back, the same as this did
before, yet Amiberry passes the same tests on a Picasso II board. Either
its board path differs from the code I read, or something there makes the
driver choose a different fill strategy. I did not chase it down.

### DIFF
--- a/src/picasso2/gd5426.rs
+++ b/src/picasso2/gd5426.rs
@@ -13,6 +13,7 @@ use crate::chipset::paula::PAULA_CLOCK_HZ;
 
 const MAX_DIM: usize = 4096;
 const BLIT_BUSY_CCK: u32 = 16;
+const BLIT_ADDR_MASK: usize = 0x1f_ffff;
 
 const SR_CURSOR_X: u8 = 0x10;
 const SR_CURSOR_Y: u8 = 0x11;
@@ -712,6 +713,13 @@ impl CirrusGd5426 {
     }
 
     fn start_blit(&mut self) {
+        // Captured before the transfer, which leaves the address registers at
+        // its end. Logging them afterwards would print the end address for an
+        // immediate blit and the start address for a deferred system-source
+        // one, so the same line would mean two different things.
+        let dst_start = self.blit_addr(0x28);
+        let src_start = self.blit_addr(0x2c);
+
         if self.gfx[0x30] & BLIT_MODE_SYSTEM_SOURCE != 0 {
             let width = self.blit_width();
             let height = self.blit_height();
@@ -729,8 +737,8 @@ impl CirrusGd5426 {
                 "picasso2: blit {}x{} dst={:#08x} src={:#08x} mode={:#04x} rop={:#04x}",
                 self.blit_width(),
                 self.blit_height(),
-                self.blit_addr(0x28),
-                self.blit_addr(0x2c),
+                dst_start,
+                src_start,
                 self.gfx[0x30],
                 self.gfx[0x32]
             );
@@ -769,6 +777,17 @@ impl CirrusGd5426 {
         usize::from(self.gfx[low])
             | (usize::from(self.gfx[low + 1]) << 8)
             | (usize::from(self.gfx[low + 2] & 0x1f) << 16)
+    }
+
+    // The counterpart, for putting an address counter back after a transfer.
+    // The registers hold 21 bits, so an address that ran off either end comes
+    // back wrapped rather than clamped; the unused top bits of the high byte
+    // are left alone.
+    fn set_blit_addr(&mut self, low: usize, addr: usize) {
+        let addr = addr & BLIT_ADDR_MASK;
+        self.gfx[low] = addr as u8;
+        self.gfx[low + 1] = (addr >> 8) as u8;
+        self.gfx[low + 2] = (self.gfx[low + 2] & !0x1f) | ((addr >> 16) as u8 & 0x1f);
     }
 
     fn blit_pixel_bytes(&self) -> usize {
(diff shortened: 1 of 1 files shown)


codewiz:
Thanks -- three of the four were real. Addressed in 4bbaba7e.

**Source counter vs. mode.** Right, and this was the one that could have
broken something that worked before. The end address is now taken from
whichever source the transfer actually read: `expand_addr` for video-source
colour expansion, which walks one byte per eight pixels and never consults
the source pitch, and the unchanged base for pattern and system-source
blits, which do not walk VRAM at all. Only a plain copy still derives it
from the destination geometry.

**Backward wrap.** Taken as offered -- `wrapping_sub`, then the low 21 bits
stored through a `set_blit_addr()` helper that leaves the unused top bits of
the high byte alone.

**Diagnostic ordering.** Correct, and it would have hidden the evidence this
PR was built on. `start_blit()` now snapshots the source and destination
before the transfer and logs those, so the line means the same thing for an
immediate blit and a deferred system-source one.

**Test coverage.** Partly fair: the existing test does fail without the
change, so it was not vacuous, but it asserted nothing about the registers.
It now checks GR28-GR2A and GR2C-GR2E after the replication blit, and a
second test, `the_source_counter_follows_whichever_source_the_mode_reads`,
covers multi-row copy, backward wrap, colour expansion, pattern and
system-source. Neutralising the per-mode branch makes it fail with
`SRC + 20` where colour expansion should leave `SRC + 3`.

`cargo fmt --check`, `cargo clippy --all-targets --all-features --locked
-- -D warnings` and `cargo test --lib picasso2` (23) are clean, and p96cts
on a `picasso2plus` board is still 121 PASS / 0 FAIL / 19 skip across all
four depths, so the source-counter rework did not cost anything end to end.

The two caveats in the description still stand: the end-address arithmetic
is what makes PicassoII.card's sequence work rather than something I can
cite from the data book, and I still cannot explain why Amiberry passes
these tests when its Cirrus core reloads the same registers on every start.

LinuxJedi (approved):

PR #578, opened 2026-08-27
ci: track driveless machines in the wasm demo smoke test

LinuxJedi:
The Browser demo publish job fails since 2a92d61b ("ui: model driveless CD machines and show CD tracks") made a floppy drive count of zero valid in the `WebEmu` constructor: the smoke test still listed `0` among the counts expected to throw, so it now dies with `invalid browser floppy drive count was accepted: 0` before publishing (see run 33084068408).

- Probe below the valid range with `-1` instead of `0` in the invalid-count loop.
- Assert the new behavior positively: `new WebEmu('A500', 'PAL', 0)` is accepted and leaves all four drives disconnected, matching the crate's `web_constructor_accepts_no_floppy_drives` unit test.


LinuxJedi:
Has been merged

PR #582, opened 2026-08-28
cd32: calibrate the drive model and CPU bus classes against real hardware

LinuxJedi:
Calibrates the CD32 drive model and two CPU bus-access classes against a real PAL CD32, measured with a purpose-built probe disc (`tools/cd32-probe`, included) and a filmed cold boot. Started from a single report — Fightin' Spirit reboot-looping with `SYS:FSCD: not enough memory available` — and ended with the emulated cold boot matching the real machine to 50 ms.

## What was wrong

Fightin' Spirit's 1.62 MB loader needs two ~720 KB contiguous chip hunks, which only exist when the extended ROM's boot show never runs. On a real CD32 with a bootable disc inserted at power-on, the show *never* runs: the drive's long spin-up holds the KS driver's first TOC transaction open, and the boot screen's one-shot `CD_CHANGESTATE` queues behind it until the disc is readable, answering "disc present" → boot. (The filmed boot shows Kickstart grey for ~11 s, no fly-in, no chime.) Copperline's drive was ready instantly and executed queued commands mid-TOC-dump, so the follow-up's reply aborted the driver's dump wait, the query answered "no disc", and the show ran — fragmenting chip RAM. The same premature-reply defect made discs inserted at the idle screen or mid-show never boot, via a TX-DMA page-carry corrupting an occasional command into a checksum error.

## Changes

- **Akiko TX DMA folds to its 256-byte page like RX** — the index registers are eight-bit and Kickstart's producer wraps packets at index `$FF` (register-trace verified). Fixes runtime disc inserts (idle screen and mid-show now boot, like a real lid close).
- **Spin-up gate on the first lead-in dump** (cold 8.8 s / warm 3.5 s; guest reset keeps the disc spinning) plus **dump-exclusive command processing** (an in-flight TOC dump completes before the next queued command acts). Together they reproduce the real cold-boot path: startup-sequence entry at 14.36 s emulated vs 14.31 s filmed, boot memory layout matching to a few hundred bytes. Fightin' Spirit boots from the unmodified disc image. `STATE_VERSION` 70.
- **Measured locate curve** replaces the WinUAE-style single-slot seek: 254 ms at +500 sectors through 1.37 s at +200 000, piecewise-linear between measured anchors, speed-scaled. The measured 2× sequential rate (150.0 sectors/s) matched the existing pacing exactly.
- **Plays into data tracks are refused** with the play-failed notification (real firmware answers io_Error 36).
- **Unmapped space bills one external cycle per 16-bit bus cycle** instead of a slow-external cycle per byte: measured 12.0 clocks per `CMP.W (A4)+`/`DBF` iteration over the empty `$A80000` window — exactly ROM pace — and 8.0 clocks for the posted write loop (one-clock overlap credit). The per-byte over-billing made exec's ROMTAG scan of that window start the whole boot show seconds late. Both discs' probe runs agree within 2 E-ticks; the previously calibrated chip read/write classes reproduced on this second machine to the tick.
- **`tools/cd32-probe`**: the measurement program and image builder, with the row map and both real-hardware result columns in its README. Every drive operation runs under a watchdog so a burned disc cannot hang the probe.
- The timing-test golden re-blessed: only its boot-phase-relative rows moved (raw beam positions and beam-race counts); every calibrated CPU-class row held within one E-clock sampling tick.

## Validation

- 3029 unit tests, 39/39 golden renders, clippy and `cargo fmt --check` clean.
- Titles: Fightin' Spirit (unpatched, to credits), Kang Fu (power-on, mid-show insert, idle-screen insert), Microcosm, KS+FMV Cannon Fodder (MPEG intro plays), and the bundled AROS ROM booting Kang Fu.
- Probe cross-check, emulated vs real: ENTRYTIM 14.36/14.31 s (full-fidelity disc) and 13.18/13.27 s (data disc); far locates within 5%; teardown block 2.019/2.016 s.


LinuxJedi:
Addressed the review in 0439e9b2:

- **Power-cycle spin-up** (src/akiko.rs): confirmed — `Bus::power_on_reset` shares the warm-reset path, so a UI/CCP power cycle skipped the cold gate. It now re-arms the gate for a mounted disc after the shared reset (`Akiko::rearm_cold_spin_up`), with a regression test covering guest-reset-keeps / power-cycle-respins / no-disc-no-op.
- **TOC pregap double-count** (make-images.py): confirmed — the Redump bins embed their own two-second pregap, so `PREGAP 0:2:0` + whole-file reads shifted every later track. The TOC now emits `FILE ... 0` + `START 0:2:0` so the embedded pregap is marked rather than duplicated.
- **Relative source paths** (make-images.py): the input path is resolved before being embedded, so the emitted TOC/CUE work from any directory including the documented `cd <outdir>` flow.
- **Void-write credit comment** (src/bus.rs): the calibration was measured with the per-byte call (two clocks per aligned word); the comment now describes that accounting instead of misstating a single-clock credit. Behavior unchanged.
- **port_init signal failure** (probe.c): halts before any device io instead of arming a bogus `Wait` mask, preserving the cannot-hang guarantee.
- Stale comments fixed (start.s owns the first hunk — build.sh and probe.c headers; amiga-mini.h names the probe with its retro32-term provenance) and the README calibration table updated to the final calibrated run.

Rebuilt PROBE, regenerated the images, re-verified the full-fidelity cue boots to the expected rows in the emulator; 3030 unit tests, 39/39 goldens, clippy/fmt clean.

PR #592, opened 2026-08-29
timing-test: accelprobe, a 020-060 accelerator probe with real A4000 columns

LinuxJedi:
A 32-row bootable probe for accelerator CPUs (68020/030/040/060), measured against the CIA E-clock like the main timing disk, with REAL columns captured on a real A4000 (50 MHz BFG9060, 25 MHz A3640 with MAPROM, 25 MHz 68030 board) recorded in the probe header alongside Copperline's predicted columns and per-row verdicts.

What it measures:
- exception round trips with the supervisor stack in chip vs fast RAM
- call, MOVEM, and CPUSHL costs
- chip-window and per-fast-region read/write rates -- the probe walks the exec MemList at boot, so one binary adapts to whichever CPU card and RAM mix is fitted (CPU-card / motherboard / Zorro III) and reports each region's base address
- a composite that replicates, primitive by primitive, the per-4K-page sequence of the SetPatch/mmu.library boot-time MMU table build, walking the largest fast region

Findings against real silicon (details and full columns in `timing-test/accelprobe.asm`, summarized in `docs/internals/cpu.md`):
- the 68030/040 scaled approximation over-bills plain 040 execution ~2-2.5x (real taken dbra ~4 clk vs ~8 modelled; mulu.w ~10 vs ~27) while under-billing its buses
- the real 030's core rows land on the same figures the real A1200 measured for the 68020 (taken dbra 6 clk, mulu ~29) -- evidence the 030 belongs on the MC68020UM tables
- fast RAM is three genuine speed classes on real silicon (CPU-card ~66ns, motherboard ~320ns, Z3 ~530ns per sequential longword) where the emulated machine bills all three at CPU speed
- the A3640/BFG9060 CPU-slot bridge prices a chip-RAM move.l read at 2.5-3.4us; the 030 board's chip path matches the emulated cost tick-exactly
- the MMU boot walk over Zorro III RAM costs 17.6us/page (040) / 12.2us/page (060) on real hardware vs 11.0/5.1 emulated -- under-billed, not over-billed

Also captured for the record: assembling vpos from separate VPOSR/VHPOSR reads can straddle the line 255->256 boundary and produce a bogus 0, which a fast CPU's poll loop hits reliably; the probe debounces with two agreeing samples (the main disk's 280 threshold dodges the hazard by construction).

All four real captures were serial streams, two runs each; the A3640 runs were byte-identical.


LinuxJedi:
All eight review comments (plus the two suppressed duplicates) addressed in 07abe10b:

- **Composite stack guard** (rows 27-29, the suppressed pair included): the rows now skip when `fast1w` cannot supply a fast stack, instead of silently running with SP at chip $18000 on a machine whose first fast region is under $30000.
- **Largest-region tracking**: the MemList capture now tracks the largest fast region across the *whole* list (`LARGLO`/`LARGHI`) so a fourth, larger region feeds the composite; the base/read/write rows keep displaying the first three, and the header says so.
- **68010 gate**: the CPU gate now requires a 68020/030/040 AttnFlags bit, so a 68010 takes the signature-only path instead of misreporting as a $20.
- **Stale text**: row-0 signature description updated to $ACCE1B02 (noting v1 captures read $ACCE1B01), the three tt-a4000 TOML headers now describe the 2M motherboard bank they configure, and the README's real-column list includes the 68030 board.

Behavior on the captured machine shapes is unchanged: the emulated 060/040/030 columns reproduce the committed predicted columns within the documented ±1-2 tick run-to-run jitter.

PR #598, opened 2026-08-29
warp boot: warp any configuration's boot until storage settles or a timestamp

LinuxJedi:
`--run` warp-launches: unpaced with muted audio until the guest loads the program. This generalizes that to configurations that boot from their own media -- a Workbench install, a hard-disk setup like AmigaVision, a CD:

- **`--warp-boot`** / `[emulation] warp_boot`: warp from power-on until the boot storage (the floppy and hard-disk activity the front-panel LEDs show) has been idle for `[emulation] warp_boot_idle` emulated seconds (default 10), then resume real-time pacing and audio. Overshoot is free -- the extra idle seconds pass at warp speed -- but the threshold must outlast the boot's longest storage-quiet stretch: a big-RAM machine's SetPatch MMU table build keeps the disk idle for seconds while the CPU walks every page of fitted RAM (the accelprobe campaign measured that walk). CD audio deliberately does not count as activity (CD *data* traffic rides the HDD LED), so a menu's background CD-DA cannot hold the warp; a boot whose storage never settles gives up after 600 emulated seconds.
- **`--warp-until SECS`** / `[emulation] warp_until`: the deterministic variant -- warp until an absolute emulated timestamp, for a setup whose boot time is known. An AmigaVision 040 machine reaches its menu in a couple of wall-clock seconds either way.

The gate (`src/warpboot.rs`) mirrors the warp launch's shape: a pure state machine over emulated time, polled once per retired frame inside the warp burst, engaged at power-on, one-shot per session, cancelled by the manual warp toggle, audio muted only while genuinely unpaced (a machine with a bridged physical drive still runs the gate at normal speed). The storage-idle condition reads `front_panel_status()`, whose HDD LED is a non-draining timestamp latch, so the gate and the status bar share the signal without stealing each other's events.

Headless captures are already unpaced end to end and drop the gate, same as the `--run` gate. Combining with `--run` (whose own warp ends at a sharper condition), or combining the two modes with each other, is a configuration error with a message saying so.

Covered by: six pure state-machine tests (`warpboot.rs`), two App wiring tests driving a live emulator (engage/poll/finish and the manual-toggle cancel), CLI/TOML validation smoke tests for both error paths, and docs in `docs/guide/configuration.md`, `docs/guide/run.md`, and `copperline.example.toml`. Full suite green, clippy/fmt clean.

### DIFF
--- a/copperline.example.toml
+++ b/copperline.example.toml
@@ -78,6 +78,23 @@ pacing_budget = "cycles"
 # it live from Warp Settings > Warp Limit or Cmd+Shift+W / Alt+Shift+W.
 # warp_speed = "max"
 
+# Warp boot: run unpaced (audio muted) from power-on until the boot storage
+# -- the floppy and hard-disk activity the front-panel LEDs show -- has been
+# idle for warp_boot_idle emulated seconds, then resume real-time pacing.
+# Same idea as --run's automatic warp launch, for any configuration that
+# boots from its own media (a Workbench install, a hard-disk setup, a CD).
+# The idle threshold must outlast the boot's longest storage-quiet stretch:
+# a big-RAM machine's SetPatch MMU table build keeps the disk idle for
+# seconds while the CPU walks every page of fitted RAM. The manual warp
+# toggle cancels the phase; headless captures are already unpaced and
+# ignore it. CD audio does not count as activity.
+# warp_boot = false
+# warp_boot_idle = 10
+
+# Warp boot until an absolute emulated timestamp instead: deterministic,
+# for a setup whose boot time is known (mutually exclusive with warp_boot).
+# warp_until = 12.0
+
 # Record rewind history from power-on, so Cmd+Z / Alt+Z (and the Rewind menu
 # item) can step the machine back through it. Off by default because it is
 # not free: every rewind_interval_frames the whole machine is serialized, and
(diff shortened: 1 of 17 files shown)


LinuxJedi:
All eight review comments addressed in 0f44b661:

- **Launcher round trip**: `MachineSetup` now carries `warp_boot`/`warp_boot_idle`/`warp_until` the way it preserves `run_ahead_frames` (no controls yet, values preserved), with a round-trip test. Worth noting separately: the `rewind`/`rewind_budget_mb`/`rewind_interval_frames` keys share this exact gap today and predate this PR -- happy to fix those in a follow-up.
- **Launcher-started machine**: `run_machine` constructs the gate from the machine's config via a new shared `warpboot::gate_from_settings` and engages it -- it is the launcher path's power-on.
- **Power lifecycle**: a powered-off start engages the gate at first power-on next to the warp launch; power-off cancels it and restores pacing, exactly like the launch gate.
- **Mode-path validation**: the two conflict checks now run before the gdb/control/benchmark dispatches, and shared config validation rejects a TOML that sets both `warp_boot` and `warp_until`, so launcher/library consumers cannot receive the invalid combination either (verified with both error paths end to end).
- **Toggle fall-through**: cancelling a gate is now a complete action (repace, OSD, return) instead of a second toggle -- this also fixes the same pre-existing bridged-drive edge in the `--run` launch cancellation, where the press was misreported as "Warp speed on".
- **Malformed warning string**: fixed (a heredoc had eaten the line continuation).
- **copperline-player**: passes the new `App::new` argument (`None`; publisher bundles keep their own warp launch), verified compiling standalone.

Full suite green including the new launcher round-trip and gate lifecycle tests; clippy/fmt clean.

PR #600, opened 2026-08-30
launcher: Warp boot rows on the A/V & Emu Emulation page

LinuxJedi:
Follow-up to #598 (merged): the warp-boot settings get launcher controls, as two rows under *Warp speed* on the *A/V & Emu* → *Emulation* page, in the page's existing cycle style:

- **Warp boot**: *Off* ↔ *Storage idle* (the boot warps until the floppy/HDD LEDs have been quiet for the idle threshold below). A `warp_until` loaded from a config file shows as its own *Until Ns* state; the modes are mutually exclusive, so one press clears it and lands on *Off* -- the panel can never produce the invalid both-set combination.
- **Warp boot idle**: preset stepper over 5s-120s (the upper presets exist because the threshold must outlast the boot's longest storage-quiet stretch -- a big-RAM machine's MMU table build keeps the disk idle for seconds).

The Emulation category also gains a UI-preview render alongside its Audio/Video/Display siblings (`COPPERLINE_UI_PREVIEW=1` → `target/ui-preview-launcher-av-emulation.png`), which is how the layout below was verified:

| Row | Value |
|---|---|
| Power on startup | Enabled |
| Realtime priority | Disabled |
| Pacing budget | Cycles |
| Warp speed | Max |
| **Warp boot** | **Off** |
| **Warp boot idle** | **10s** |
| WHDLoad | Enabled |

Row cycling and the round trip through `RawConfig` are covered in the launcher tests, including the `warp_until`-clearing press. Docs: the `configuration.md` warp-boot bullet now names the launcher rows. Full suite green, clippy/fmt clean.

(Housekeeping: a stray push had resurrected the deleted `feat/warp-boot` branch after #598 merged; it has been deleted again and this commit re-based onto main.)

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -434,7 +434,10 @@ carried no information.)
   are mutually exclusive, `--warp-boot`/`--warp-until` set them from the
   command line, the manual warp toggle cancels the phase, headless captures
   (already unpaced end to end) ignore both, and `--run` (whose own warp
-  launch ends at a sharper condition) cannot combine with them.
+  launch ends at a sharper condition) cannot combine with them. The
+  launcher's *Warp boot* and *Warp boot idle* rows (*A/V & Emu*,
+  *Emulation*) edit the storage-idle mode; a `warp_until` loaded from a
+  config shows there as its own *Until Ns* state, which one press clears.
 - `rewind = true` records rewind history from power-on, so `Cmd+Z` / `Alt+Z`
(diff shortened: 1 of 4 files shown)


LinuxJedi:
Both label sites fixed in the follow-up commit: a `format_secs` helper (same shape as `format_mhz`) keeps whole seconds terse ("10s") and fractional configured values exact ("Until 12.5s") on the *Until* state and the idle-threshold row alike; the round-trip test now pins the 12.5 label.

PR #616, opened 2026-09-01
control: MCP server mode for copperline-ctl

LinuxJedi:
An MCP server mode for `copperline-ctl`, so an agent in Claude Code, Cursor, or any other MCP client drives a live machine through the control protocol as tools instead of a REPL.

## What changes

- **`copperline-ctl --mcp [--info FILE | --connect ADDR --token TOKEN]`** (`src/control/mcp.rs`): an MCP server over stdio. Every control-protocol method is a tool named after it with dots replaced by underscores (`warp.get` -> `warp_get`); with connection arguments it attaches at startup, without them the agent launches or attaches a session through the session tools. Claude Code: `claude mcp add copperline -- copperline-ctl --mcp`, or an `.mcp.json` entry.
- **Tool catalogue** (`src/control/catalogue.rs`): one entry per method with a one-paragraph description, a JSON Schema for its params (required fields, enums, integer-or-hex-string addresses, bounds from the parser) and an example. Tests pin it to `exec::parse_method` both ways: every string-literal arm of the parser has an entry, every entry is a method the parser or a driver answers, every example parses. `hello`/`auth` stay the bridge's own handshake.
- **Bridge** (`src/control/bridge.rs`): one authenticated connection; a reader thread owns the receive side, routes replies by id and queues `event.*` notifications (bounded to 1024, drops counted) so subscriptions keep collecting while a resume verb's reply is outstanding. It also launches a headless emulator for `session_launch` (`copperline --control :0 --control-info TMP --noaudio` plus the caller's flags; binary from the argument, `$COPPERLINE_BIN`, the sibling of `copperline-ctl`, or the PATH), waits for the endpoint and sends the process's output to a log file named in the result.
- **Session tools**: `session_launch {config, model, run, whdload, factory, args, binary, cwd, timeout_ms}`, `session_attach {info_file} | {listen, token}`, `session_status`, `session_close`. One session at a time; an emulator this server launched is shut down on close and on stdin EOF, killed after 3 s if it does not exit.
- **Event tools**: `events_next {timeout_ms}` (blocks for the next queued notification or the timeout) and `events_drain`.
- **`wait_ms`** on the resume verbs (`continue`, `run_until`, `step`, `step_over`, `step_out`, `step_copper`, `step_frame`): both servers reply to a resume with the eventual stop event and MCP handles one request at a time, so a `continue` with no breakpoint would block the server for good. On expiry the bridge sends `pause` and returns the stop that produces, marked `bridge.paused_after_ms`; a stop that arrives in time is returned as it is.
- **`capture_screenshot`** returns the PNG as an MCP image content block alongside the text result, so the model can look at the screen; with no `path` the file is temporary and deleted after reading.
- `ctl-bin` now depends on the `control` feature (the bin links the library's `mcp` module); the bin itself stays a thin shim on std + serde_json, no rmcp/tokio.
- Docs: an "MCP server" section in `docs/debugger/control.md` (command line, Claude Code setup, name mapping, session tools, `wait_ms`, events, screenshots, protocol subset); a paragraph in AGENTS.md's Live control section; README's debugging bullet; `docs/index.md`.

## Design

- **Protocol subset, hand-rolled.** Newline-delimited JSON-RPC 2.0 on stdin/stdout: `initialize` (protocolVersion echoed when it is 2025-06-18, 2025-03-26 or 2024-11-05, else the latest; capabilities `tools.listChanged: false`; serverInfo `copperline` / CARGO_PKG_VERSION; an `instructions` workflow summary), `notifications/initialized` and other notifications ignored, `ping`, `tools/list`, `tools/call`. Unknown methods get -32601, malformed messages -32600/-32700; a tool failure, including a control-protocol error reply, is a `tools/call` result with `isError` and the code and message in the text, never a JSON-RPC error, so the model can read it and adapt. Stdout carries protocol messages only; diagnostics go to stderr, and the launched emulator's stdout/stderr go to its log file rather than inheriting ours.
- **No lock across a blocking wait.** The request thread does the control-protocol round trip itself and waits on a condition variable for its id; the reader thread only takes the state mutex to store a reply or push an event. A closed connection fails every waiter with the reason and is reported by `session_status`.
- **Name mapping is a table lookup**, not a string edit: `run_until` and `step_over` carry underscores of their own, so the reverse map goes through the catalogue.
- The `wait_ms` handling lives in the bridge, not the servers: the headless and windowed servers already service `pause` mid-run at a quantum boundary and answer both ids with the stop position, so no server change was needed.
- `headless.rs` gains a `#[cfg(test)]` helper that serves one session on a background thread, so the MCP tests run against a real headless session (the bundled test ROM) from the test thread.

## Verification

- `cargo test`: 3175 lib tests passed (new: catalogue coverage/round-trip/example tests; bridge routing, timeout, bounded queue, lost connection, launch failure reports; MCP initialize handshake, malformed/unknown requests, tools/list contents, `status` through the bridge, unknown tool and invalid-params as `isError` results, hex address round trip, `continue` with `wait_ms` pausing a free-running machine and a breakpoint stop arriving unmarked, event queue flow, screenshot image blocks, one-session-at-a-time, `shutdown` closing the session, `serve` over in-memory reader/writer). `cargo clippy --all-targets` and `cargo fmt --check` clean.
- `cargo test --release --test mcp_stdio -- --ignored`: passes (1.05 s): the built `copperline-ctl --mcp` driven over its stdio launches the built emulator on the bundled AROS ROM, runs to frame 5 with `wait_ms`, returns a screenshot image block, closes the session, exits on EOF, and the emulator pid is gone afterwards.
- By hand, against the release binaries:

  ```sh
  ./target/release/copperline --factory --noaudio --control :0 --control-info /tmp/x.json &
  printf '%s\n' \
    '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-06-18","capabilities":{},"clientInfo":{"name":"shell","version":"0"}}}' \
    '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
    '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' \
    '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"status"}}' \
    '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"run_until","arguments":{"seconds":2,"wait_ms":20000}}}' \
    '{"jsonrpc":"2.0","id":5,"method":"tools/call","params":{"name":"capture_screenshot"}}' \
    '{"jsonrpc":"2.0","id":6,"method":"tools/call","params":{"name":"session_status"}}' \
    | ./target/release/copperline-ctl --mcp --info /tmp/x.json
  ```

  `initialize` answers with the capabilities and instructions, `tools/list` lists 80 tools (6 bridge tools + 74 methods), `status` returns the JSON text block with `frame: 0`, `run_until` returns `{"reason": "target", "frame": 99, "seconds": 2.0000008, ...}`, `capture_screenshot` returns a text block plus an `image/png` block, nothing but protocol lines reaches stdout, and the emulator log shows the clean detach.

  ```sh
  printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"shutdown"}}' \
    | ./target/release/copperline-ctl --mcp --info /tmp/x.json
  ```

  ends that emulator (`session.closed: true`, the process exits).

  ```sh
  printf '%s\n' \
    '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"session_launch","arguments":{"factory":true,"model":"A500","args":["/path/to/KICK13.ROM"]}}}' \
    '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"run_until","arguments":{"seconds":8,"wait_ms":60000}}}' \
    '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"capture_screenshot","arguments":{"path":"/tmp/kick13.png"}}}' \
    '{"jsonrpc":"2.0","id":4,"method":"tools/call","params":{"name":"session_close"}}' \
    | ./target/release/copperline-ctl --mcp
  ```

  launches the sibling `copperline` (the result carries the pid, address, log path and full command line), runs to 8 s, writes the Kickstart 1.3 insert-disk screen (also attached as the image block, checked by eye), and `session_close` reports `killed: false`: the emulator exited on `shutdown` by itself. The same without a ROM boots the bundled AROS.
- Not run: the windowed `--control-gui` attach (it would open a window on the dev host); the bridge speaks the same wire format to both servers and the windowed drain already answers `pause` mid-run.

## Judgment calls

- `wait_ms` is offered on every resume verb, not only `continue`/`run_until`/`step_frame`: `step_out` and `step_over` run to the instruction budget when the routine never returns, and a uniform rule is easier to explain than an exception.
- The whole catalogue is exposed, 80 tools: the brief asked for the full protocol, and a client that wants fewer can filter by name. `tools/list` descriptions say what each returns and whether it blocks, so the model does not need the reference.
- `shutdown` is a tool (it is a documented method); after it succeeds the bridge closes the session and reports so in the result, since the emulator is on its way out either way.
- The end-to-end stdio test is `#[ignore]` like the rest of `tests/`: it spawns two processes and runs the emulator, even though it needs no local assets.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -159,6 +159,14 @@ copperline-ctl --info /tmp/ccp.json --repl
 events.subscribe {"events":["frame","serial","interrupt","media"],"frame_interval":50}
 ```
 
+An MCP-capable agent (Claude Code, Cursor, ...) gets the same protocol as
+tools: `claude mcp add copperline -- copperline-ctl --mcp` (or an `.mcp.json`
+entry), then `session_launch {"config": "my.toml"}` or `session_attach
+{"info_file": "/tmp/ccp.json"}`, `break_add`, `continue {"wait_ms": 5000}`
+(the bridge pauses the machine when the wait expires), `capture_screenshot`
+(returned as an image), `events_next`. Tool names are the method names with
+dots as underscores; see the "MCP server" section of the reference.
+
 The event streams are bounded, so check their drop counts when observations
 must be complete. The same session can use `trace.start`/`trace.stop` and
 `waveform.start`/`waveform.stop` to bracket file-backed diagnostics at runtime
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,8 +134,10 @@ net-nat = ["dep:smoltcp"]
 # copperline-net-helper when an unprivileged open is denied), macOS uses the
 # system libpcap, and Windows loads an installed Npcap at runtime.
 net-bridge = ["dep:pcap", "dep:libloading"]
-# The `copperline-ctl` command-line client for the control protocol.
-ctl-bin = ["dep:serde_json"]
+# The `copperline-ctl` command-line client for the control protocol, and
+# its `--mcp` mode (src/control/mcp.rs), an MCP server over stdio that
+# exposes the protocol as tools for coding agents.
+ctl-bin = ["dep:serde_json", "control"]
 # The `copperline-import-uae` converter for WinUAE/Amiberry/FS-UAE
 # configuration files.
 import-uae-bin = ["dep:toml_edit"]
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ It boots out of the box with the bundled open-source AROS Kickstart replacement,
 - **Storage and media**: Floppy disk images (ADF, ADZ, DMS, IPF, SCP), physical floppy drives via Greaseweazle (FluxBridge), IDE (Gayle/A4000), SCSI (A2091, A3000, A4091), CD-ROM, and host directory filesystem mounting.
 - **Audio and video**: 4-channel Paula audio, RTG graphics cards (Picasso II/II+, Z3660), host MIDI in/out bridging, and built-in Roland MT-32 and General MIDI synthesis.
 - **Expansion and networking**: Zorro II/III autoconfig, A2065 Ethernet, host-backed bsdsocket.library, and sandboxed WebAssembly expansion plugins.
-- **Debugging and automation**: In-window interactive debugger with reverse stepping, signal waveform export (VCD), GDB remote stub, deterministic save states, headless scripting, and a JSON-RPC control protocol (`copperline-ctl`).
+- **Debugging and automation**: In-window interactive debugger with reverse stepping, signal waveform export (VCD), GDB remote stub, deterministic save states, headless scripting, and a JSON-RPC control protocol (`copperline-ctl`, also served as MCP tools for coding agents by `copperline-ctl --mcp`).
 - **Direct launching**: Boot directly into WHDLoad game packages (`--whdload`) or host-built Amiga executables (`--run`), with a WinUAE-compatible `uaelib` trap so cross-compiler workflows can toggle warp, log, and register debug resources from the guest.
 - **WebAssembly build**: Run directly in modern web browsers at [copperline.dev/try](https://copperline.dev/try/).
 
--- a/docs/debugger/control.md
+++ b/docs/debugger/control.md
@@ -41,6 +41,121 @@ copperline-ctl --info /tmp/ccp.json continue
 copperline-ctl --info /tmp/ccp.json --repl
 ```
 
+(mcp-server)=
+## MCP server
+
+`copperline-ctl --mcp` turns the same client into a
+[Model Context Protocol](https://modelcontextprotocol.io) server over stdio,
+so an agent in Claude Code, Cursor, or any other MCP client drives a live
+machine through tools instead of a REPL. Every control-protocol method is a
+tool, and a few bridge-owned tools manage the session.
+
+```sh
+# Unattached: the agent launches or attaches a session with the session tools.
+copperline-ctl --mcp
+
+# Attached at startup to a running control server:
+copperline-ctl --mcp --info /tmp/ccp.json
+copperline-ctl --mcp --connect 127.0.0.1:7710 --token HEX
+```
+
+Claude Code registers it with one command:
+
+```sh
+claude mcp add copperline -- copperline-ctl --mcp
+```
+
+or, checked into a project, `.mcp.json`:
+
+```json
+{
+  "mcpServers": {
+    "copperline": {
+      "command": "copperline-ctl",
+      "args": ["--mcp"],
+      "env": {"COPPERLINE_BIN": "/path/to/copperline"}
+    }
+  }
+}
+```
+
+`initialize` returns an `instructions` summary of the workflow, and
+`tools/list` carries a description, a JSON Schema and the parameter
+conventions for every tool, so the agent needs nothing else from this
+chapter.
+
+### Tool names
+
+MCP tool names allow only `[a-zA-Z0-9_-]`, so a method's tool is the method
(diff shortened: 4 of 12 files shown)


LinuxJedi:
Review points addressed in 2557da27, c6c92d3b and 41e73d0a:

1. **Launched emulator's log permissions** (`bridge.rs`): the log is now created through the same owner-only options as the `--control-info` file, factored into `control::owner_only_create()` (0600 on unix, the platform default elsewhere), with the mode set as the file is created and `create_new` so a file or symlink planted under the predictable temp name is refused rather than written through. The fake-emulator launch test asserts the log's mode; the e2e run's log came out `-rw-------`.

2. **JSON-RPC envelope validation** (`mcp.rs`): `handle_message` requires `"jsonrpc": "2.0"`, accepts only string or integer ids (null, booleans, floats, objects and arrays get `-32600` with a null id), rejects a non-string `method`, and answers a message with no method and no `result`/`error` with `-32600` whether or not it has an id. Valid notifications with unknown methods and stray responses are still ignored. New test `envelopes_must_be_json_rpc_2_with_a_string_or_integer_id`; the docs' protocol-subset paragraph now states the rules.

3. **Pending-map leak after a `wait_ms` expiry**: the bridge gained `forget(id)` (drops the pending entry, so the reader discards a late reply instead of parking it) and `outstanding()`. `resume_with_wait` waits a 50 ms grace for the pause's own reply (the server writes it right behind the stop), forgets its id either way, and forgets the resume itself if it is still unanswered after the 30 s pause grace; the timeout guarantee is unchanged. Tests: `a_late_pause_reply_leaves_nothing_pending` (a scripted server holds the pause reply back until the next request; the map stays empty across three timed resumes) and `a_forgotten_request_s_late_reply_is_dropped` on the bridge. The scripted-server helpers moved to `bridge::test_support` for the MCP tests.

4. **Non-string screenshot `path`**: forwarded as given, so the method's own invalid-params error (`bad path`) comes back as an `isError` result instead of a temporary screenshot. Covered in `screenshots_come_back_as_images`.

5. **Relative screenshot `path`**: an explicit path is made absolute against `copperline-ctl`'s working directory (`std::path::absolute`) before it is forwarded, and the text result carries that absolute path; a PNG the emulator wrote but the bridge cannot read back is now an `isError` result rather than a success with a note. Test added for the relative case; the Screenshots section of the docs describes both.

Verified: `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` (3178 lib tests, 3 new), and `cargo test --release --test mcp_stdio -- --ignored` (passes, 1.02 s).

LinuxJedi:
CI: the Linux unit-test job failed on `launch_reports_a_process_that_exits_without_announcing` with `File exists` on the launch log. The two launch tests run in parallel threads of one process, and both picked the same `copperline-mcp-PID-MILLIS` name inside one millisecond, which the new `create_new` guard correctly refused. 01ba7fbc adds a per-process sequence number to the name, skips a name that already exists, and removes the empty log when the spawn itself fails; `launches_in_the_same_millisecond_get_distinct_private_files` pins it.

PR #617, opened 2026-09-01
debugger: pin headless observation as timeline-transparent, drop the crash alarm from save states

LinuxJedi:
The headless debugger (`COPPERLINE_DBG_*`) is pinned as timeline-transparent, the one way a debugged run silently diverged from a plain one is now announced, and the one byte by which a resumed AGA save state differed from the uninterrupted run's is gone.

## What changes

A note from an August investigation claimed that arming `COPPERLINE_DBG_*` knobs shifted the emulated timeline before their window even opened, so nothing could be instrumented without changing what was being measured. Re-measured knob by knob on three workloads (KS1.3 A500 with a disk-loading game, KS3.1 A1200/68020 with an AGA loader, the bundled AROS ROM): save-state and screenshot hashes at 25 s are byte-identical with WATCH, BREAK (139 hits), CATCH, CATCHALERT, TRACE (7169 lines), DUMP, SHOT (three mid-frame renders), IRQ, MEMW, FC, SPREN, DSKLEN, BLIT, FRAMESTATE, COPPER, RR and RWATCH armed. The August divergence was the control-protocol quantum bug fixed in d70d0b16. Two real findings remained:

- **`[cpu] jit` runs drop to the precise loop when any hook is armed** (`jit_fast_path_ok`), so the debugged run is the precise-timing run and not the JIT run, and nothing said so. `M68kMachine::note_jit_debug_fallback` now logs a warning when JIT is enabled on a machine whose hooks force the fallback, at `set_jit_enabled` and at `arm_headless_debugger`.
- **Resumed AGA states differed from the uninterrupted run's by one byte.** `Bus::diag_lowmem_blit`, the `COPPERLINE_DIAG_CRASH`-only alarm latched by a blit whose D destination lands in low memory (Kickstart 3.1 AGA game loaders do this), was serialized but cleared by the loader; on an undiagnosed run it stays latched for good, so `RWATCH` (which snapshots and restores around its backward query) and `--load-state` produced state files one byte off. It is host diagnostics, so it is `#[serde(skip)]` now (state version 74). The timeline was never affected: the blob diff was that single byte.

Also: `Debugger::new(addr_mask)` and `M68kMachine::arm_headless_debugger` arm the environment-style debugger programmatically, which is what lets a test hold the guarantee without the `envcfg` snapshot.

## Verification

- New `tests/debugger_transparency.rs` (release-only, asset-free): two machines boot the bundled ROM side by side, one armed with a watch over the vector/ExecBase area, a breakpoint on the reset vector, the VERTB exception catch, Alert() catching, a memory dump on every hit and an unbounded instruction trace; fingerprints (colour clock, frame, PC, rendered frame, chip RAM) are compared after every one of 756 frames. Passes in 8.6 s.
- `tests/savestate_roundtrip.rs` now also compares the resumed run's state file at the last checkpoint byte for byte with the uninterrupted run's. Passes.
- The bundled-ROM boot does not trip the alarm, so the CLI check that found it was re-run against this build: an A1200 KS3.1 workload with `--rtc-time` pinned, `--save-state-after 15` then `--load-state` plus `--save-state-after 25`, against a plain `--save-state-after 25`. One byte apart before the change; identical after (both `d4d872c6...`). The same workload under `COPPERLINE_DBG_RWATCH=0x400` matches two baseline runs on both the state hash and the screenshot hash.
- `src/bus/tests.rs`: serializing a bus with the alarm latched and clear gives identical bytes.
- `cargo test --lib` (3149 passed), `cargo clippy --all-targets --all-features --locked -- -D warnings`, `cargo fmt --check`, `git diff --check` all clean.
- Probe recipe, for the record: `--save-state-after 25 --screenshot-after 25` under each knob with `--rtc-time` pinned, hashes compared against two baseline runs; the differing field was found by tree-diffing `serde_json::to_value` of both buses.

## Docs

`docs/debugger/headless.md` gains a "Timeline transparency" section stating the guarantee, the JIT exception and the model knobs that do move a timeline by design; `docs/internals/savestate.md` lists the alarm among the skipped host diagnostics; AGENTS.md says the knobs are safe to layer onto one run.

## Judgment calls

- The alarm is skipped rather than made to survive a load: it is consumed only under `COPPERLINE_DIAG_CRASH`, and a resumed run should not dump a crash ring for a blit that happened before the state was taken.
- State version bump: removing a field changes the layout, so this branch moves `STATE_VERSION` to 74. The HRTmon cartridge branch (`feat/hrtmon-cartridge`) also moves it to 74; whichever of the two merges second must move to 75.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -176,7 +176,11 @@ debuggers.
   environment variables: breakpoints, watchpoints, instruction traces,
   Copper-list dumps, per-hit screenshots (`docs/debugger/headless.md`).
   All `COPPERLINE_*` variables are snapshotted at startup and cannot change
-  at runtime.
+  at runtime. Arming them is timeline-transparent: the debugged run
+  retires the same instructions at the same colour clocks as the plain
+  run, so instruments can be added one at a time to the same run (the one
+  exception is `[cpu] jit`, which falls back to precise timing while any
+  hook is armed and says so in the log).
 - `--waveform out.vcd` with `--wave-trigger`/`--wave-duration` exports a VCD
   chip-signal trace for GTKWave (`docs/debugger/waveform.md`).
 - `--benchmark-until SECS` measures host-CPU cost of the deterministic
--- a/docs/debugger/headless.md
+++ b/docs/debugger/headless.md
@@ -18,6 +18,34 @@ COPPERLINE_DBG_SHOT=/tmp/hit \
 
 Addresses are specified in hexadecimal (with or without `0x` or `$` prefixes).
 
+## Timeline transparency
+
+The headless debugger is a pure observer. Its hooks read machine state
+through side-effect-free peeks, never bill chip-bus or CPU time, and write
+only to the host log, so arming any combination of the variables below
+leaves the emulated timeline byte-identical to an undebugged run: the same
+instructions retire at the same colour clocks, chip RAM holds the same
+bytes, and every frame renders the same. An investigation can therefore
+start from a plain `--screenshot-after` run and add breakpoints, watches,
+catches, traces, and dumps one at a time, knowing that each new
+instrument sees the same run the previous one did. The guarantee is held
+by `tests/debugger_transparency.rs`, which boots the bundled ROM twice,
+once with every hook armed, and compares the two machines after every
+frame.
+
+Two things do move a timeline, and both are documented behaviour rather
+than instrumentation cost:
+
+- A `[cpu] jit` machine (`--jit`) runs the batch/trace core, which bills
+  time in coarse batches instead of per instruction. Any armed debug or
+  diagnostic hook needs the per-instruction loop, so the machine drops back
+  to it for the whole run and logs a warning at start-up: the debugged run
+  is then the precise-timing run, not the JIT run. Compare debugged and
+  undebugged runs with JIT off.
+- Model knobs are not observation: `COPPERLINE_IRQ_LATENCY_CCK`,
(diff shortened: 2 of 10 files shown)


LinuxJedi:
Review points addressed in the follow-up commit:

- **MEMW missing from the JIT fallback:** `jit_fast_path_ok` now goes through a dedicated `debug_hooks_armed` predicate that includes `dbg_memw_addr`, so a JIT machine with `COPPERLINE_DBG_MEMW` armed takes the precise loop (where `debug_check_memw` runs) and the warning fires. Unit test `jit_memw_hook_forces_the_precise_loop`.
- **False warning on an FPU-less 68040:** the warning keys off `debug_hooks_armed`, which excludes the FPU shim; the shim still keeps that model on the precise loop. Unit test `fpu_less_68040_shim_is_not_a_debug_hook`.
- **Transparency test scope:** `DiagnosticHooks` / `M68kMachine::arm_diagnostic_hooks` arm the CPU-level observers programmatically (MEMW, FC, SPREN, IRQ window, IPL split, PC histogram, crash ring); the test now arms all of them alongside the debugger struct and still compares 756 frames byte-identical (12 s). The bus-level log-only knobs (CIA, DSKLEN, BLIT, FRAMESTATE) read the environment at their log sites and have no programmatic switch; the docs and the test header now say so and point at the CLI probe on record.

Verified: `cargo test --lib` 3151 passed, `cargo test --release --test debugger_transparency` passes, clippy `--all-targets` and fmt clean.

PR #618, opened 2026-09-01
hrtmon: freezer cartridge with the bundled HRTMon 2.39 monitor

LinuxJedi:
An Action Replay-style freezer cartridge with the bundled HRTMon 2.39 monitor in it: `[cartridge] model = "hrtmon"` (or `--cartridge hrtmon`) fits it, and the button -- *Tools > Freeze (HRTMon)*, `Cmd+Shift+B` / `Alt+Shift+B`, `--freeze-after SECS`, or the control protocol's `cartridge.freeze` -- drops whatever is running into the monitor on its own screen, from where the monitor's `x` returns to the program exactly where it was.

## What changes

- **The image.** `assets/hrtmon/hrtmon.rom` is HRTMon 2.39 (Alain Malek's monitor, maintained by Bert Jahn and contributors, GPL-2.0-or-later) assembled from the upstream source at https://github.com/wepl/hrtmon commit `3af8d5105f1e01ecc7961475568826e564995068` by `hrtmon-rom/build.sh` in its UAE cartridge configuration (`UAE=1`, `CARTRIDGE=1`, `SAVE_CUSTOM=1`):

  ```sh
  VASM=/opt/amiga/bin/vasmm68k_mot NDK_INCLUDE=/path/to/NDK3.2/Include_I ./hrtmon-rom/build.sh bundle
  ```

  vasm 2.0b with upstream's own flag set; the `$VER` date is pinned to the commit's so the image reproduces byte for byte (199,300 bytes, SHA-256 `184c9b12e7e83749d817b1c2692e5a86dffb84dc1bcda0d8c752650dfafe3d74`). Nothing from any other emulator is used. `hrtmon-rom/README.md` has the provenance and the five-line build-time patch; the license text sits beside the image; the formula, the Flatpak manifest and the AppImage and DMG builds ship it; `CREDITS.md`, `RELEASE.md` and the README credit it.
- **The model** (`src/cartridge.rs`, `Bus`, `CpuBus`): a 1 MiB bank at `$A10000` present from power-on (plain read/write memory; the monitor's variables, stack and screen live there); the configuration block at the head of the image written from the fitted boards at attach and at every reset (chipset, video standard, IDE port, CD32, chip RAM size; `entered` cleared); a shadow of every custom-register write (CPU and Copper) and driven read plus each CIA register's last byte, copied into the bank on a freeze at the offsets the `SAVE_CUSTOM` build reads (`$A9F000`, `$A9E000`, `$A9D000`); and the freeze itself: the level-7 autovector under the current VBR pointed at the header's `bra.w monitor`, and a non-maskable interrupt the CPU takes at the next boundary whatever the SR mask, consumed at its acknowledge so the monitor is entered exactly once per press. All of it is machine state (`STATE_VERSION` 75).
- **The triggers**: `[cartridge] model` / `rom` and `--cartridge`; a *Freezer cartridge* row on the launcher's System tab; the menu row and hotkey; `--freeze-after` and the `freeze-after` script directive (scheduled and recorded like the other inputs); `cartridge.get` / `cartridge.freeze` on the control protocol; a journaled press for reverse-debug replay.
- **Docs**: `docs/guide/configuration.md` (`[cartridge]`, `--cartridge`), `docs/guide/ui.md`, `docs/guide/headless.md`, `docs/debugger/control.md`, `docs/internals/peripherals.md` (the model), `copperline.example.toml`, README, `tests/README.md`.

## Design

- The model follows what HRTMon's UAE build expects of its host, which is also how WinUAE hosts it, so the same image works in both. Nothing here is Amiga hardware: a real Action Replay sits on the expansion bus and snoops the chip-register writes itself, which is what the shadows stand in for.
- Level 7 is the 68000's non-maskable level, recognised on its transition and not its level: `pending_irq_level` and the IPL sample report 7 while a press is armed, the core takes it under any mask, and `interrupt_acknowledge(7)` consumes it, so the RTE the monitor exits through does not re-enter it. The vector is written through the RAM-only path DMA masters use, so a VBR moved into fast RAM works and a VBR in ROM refuses the press with a reason.
- The bank is decoded after RAM, ROM and the extended ROM at external-bus cost like Zorro RAM and is cacheable; the data cache is dropped at the acknowledge in case the vector slot sat under a stale line.
- The cartridge is attached after the IDE interfaces and the CD controllers so the configuration block reports them. With an IDE port and no drive, the monitor's first entry spends up to 1.6 s per drive in its CIA-TOD-timed identify probe before it paints; that is the monitor's own behaviour.

## Verification

- `tests/hrtmon_freeze.rs` (new; release-only, asset-free: bundled AROS plus the bundled image): boots to 14 s, freezes through the library, and at 17 s checks the `entered` flag the monitor sets in its block, that the dominant screen colour is the block's COLOR00 (`$005A`), and that a state taken inside the monitor resumes inside it; a second test runs the real binary with `--factory --noaudio --cartridge hrtmon --freeze-after 14 --save-state-after 17 --screenshot-after 17`, loads the state and decodes the PNG. Both pass (`cargo test --release --test hrtmon_freeze`, 4 s).
- Unit tests, 17 new: `cpu::tests::cartridge_freeze_delivers_one_level7_interrupt_through_the_written_vector` (a stub handler at +12 under an SR mask of 7: taken once, RTE restores the mask, not re-taken until the next press); `bus::tests::cartridge_config_block_describes_the_machine_and_is_rewritten_by_a_reset` and `cartridge_shadows_follow_custom_and_cia_writes_into_the_bank_on_a_freeze`; `savestate::tests::cartridge_bank_and_shadows_travel_in_the_state`; `control::exec::tests::{parse_cartridge_methods_map_to_core_ops, cartridge_methods_report_not_found_without_a_cartridge, cartridge_get_describes_the_bank_and_freeze_arms_the_interrupt}`; three `config::tests::cartridge_*`; `inputrec::tests::a_cartridge_freeze_is_recorded_as_a_freeze_after_line`; `video::launcher::tests::the_freezer_cartridge_row_cycles_and_round_trips`; `video::menu::tests::the_freeze_row_is_offered_only_with_a_cartridge_fitted`; and the cartridge module's own four.
- Headless, by hand: `--factory --noaudio --cartridge hrtmon --freeze-after 14 --screenshot-after 17` on the default A500 (68000), with `--cpu 68020`, and `--model A1200 --fast 8M` (screenshot at 21 s, past the IDE probe) all show the monitor: registers, the AROS idle-loop PC, `MC68000`/`MC68020`, `Hrt:[$00A10000]`. The first build of the image took an illegal instruction on the 68000 at `$A10110` (`TST.B (d16,PC)`), which is what led to the `OPT a-` patch below.
- `cargo test` (3166 passed), `cargo clippy --all-targets --all-features --locked -- -D warnings`, `cargo fmt --check`, `git diff --check` all clean.

## Judgment calls

- `OPT a-` rather than lowering the vasm block's `MC68030` to `MC68000`: the source's run-time-guarded `movec` sites between `mon_install` and its first inline `MC68010` are not wrapped, so a 68000 CPU directive refuses to assemble them; switching off only the optimisation that produced the 68020+ form leaves every instruction as upstream wrote it. Upstream's own vasm build carries the same optimisation.
- WinUAE also copies the first KiB of chip RAM to `$A9C000` on a freeze; HRTMon 2.39 never reads it, so it is not modelled.
- The shortcuts panel gained its 25th row by trimming the gap above its notes and its bottom margin by two pixels each, rather than dropping a row.
- State version: this branch moves `STATE_VERSION` to 75, on top of the 74 that PR #617 (`fix/dbg-knobs-timeline`) landed first.


LinuxJedi:
Addressed the CI failures and both review points.

**CI: "Headless core and browser build" and "Publisher-kit player crate"** (cd51058b)
- `anyhow::Context` was imported into `src/emulator.rs` only under the `cd32-fmv` feature while the cartridge attach used `with_context` unconditionally, so `cargo check --no-default-features` and `--features frontend,cpu-jit` failed with E0599. The import is now unconditional.
- The same job also exposed that `App::new` grew the `freeze_after` schedule without the player crate's call site following; `crates/copperline-player` now passes an empty schedule (a bundle fits no freezer cartridge).
- Reproduced before and re-checked after with the jobs' own commands: `cargo check --no-default-features --locked`, `--features bench-bin`, `--features frontend,cpu-jit`, and `cargo check --locked` plus `cargo fmt --check` inside `crates/copperline-player`.

**Windows zip does not ship `assets/hrtmon`** (d294c275, plus a pending workflow commit)
- `packaging/windows/build-zip.ps1` now stages `hrtmon\hrtmon.rom`, `README.md` and `LICENSE` next to the exe, the same way as the aros/fmv/a4091/a2091/lide folders. The dmg, AppImage and flatpak manifests were re-checked and already stage all three files.
- Checked locally against a copy of the release binary in the zip's sibling-folder layout: `--cartridge hrtmon --freeze-after 14 --screenshot-after 17` exits 0 with `hrtmon\` staged and fails at config resolution ("bundled HRTMon image was not found") without it.
- The zip smoke test in `.github/workflows/windows.yml` gains that second run (after the existing bare boot: assert `hrtmon\hrtmon.rom` is present, then boot with `--cartridge hrtmon --freeze-after 14 --screenshot-after 17` and require a clean exit and a non-empty PNG). That commit is ready on the branch locally but could not be pushed from this session: the stored GitHub credential lacks the `workflow` scope, which GitHub requires for any push touching `.github/workflows/`. It will follow with a workflow-scoped push.
- `RELEASE.md`'s Windows section lists the sibling ROM folders.

**`--cartridge none` leaves an inherited `[cartridge] rom`** (699eaf19)
- `ConfigOverrides::apply_to` clears `raw.cartridge.rom` when the override parses as no cartridge, so `--cartridge none` over a file with `model = "hrtmon"` plus `rom = "..."` unfits the cartridge instead of tripping "rom applies only to a fitted cartridge". A model override keeps the file's image, and a misspelt model leaves it alone for the parser to report.
- New test `cartridge_none_override_clears_an_inherited_image` layers all three cases over a config file with model + rom.

Verified: `cargo fmt --check`, `cargo clippy --all-targets --all-features --locked -- -D warnings`, the full `cargo test` unit suite (3167 lib tests), the four CI feature-set checks above, and `cargo test --release --test hrtmon_freeze` (both tests pass).

LinuxJedi:
Merged main (ebc07399): #617 landed its `STATE_VERSION` 74 first, so the cartridge entry is now version 75 and the PR description says so. Unit suite, clippy, fmt and the release `hrtmon_freeze`, `savestate_roundtrip` and `debugger_transparency` tests pass on the merge.

LinuxJedi:
Merged main again (e0e7ba99): #616, #619, #621 and #623 landed. Textual conflicts were keep-both in README, `docs/debugger/control.md` and the `copper.list` parser arm. The MCP catalogue from #616 now lists `cartridge.get` and `cartridge.freeze`, and also `profile.start/stop/status` from #621, which main is missing (its `every_parser_arm_has_a_catalogue_entry` test fails on main today). Unit suite, clippy, fmt and the release `hrtmon_freeze`, `savestate_roundtrip`, `debugger_transparency` and `mcp_stdio` tests pass on the merge.

PR #653, opened 2026-09-06
Add browser rollback netplay over WebRTC

LinuxJedi:
Browser users can now start two-player rollback sessions through **Controls → Netplay**. The host shares a WebRTC offer code, the joining player returns an answer, and both pages cold-boot after their initial machines match. Each player owns one Amiga controller port; keyboard input is shared.

Reuse the existing rollback engine and wire protocol through bounded packet transports, retaining native UDP support. Browser setup supports optional STUN discovery, cancellation and fresh-session restart. Machine/media/state and floppy sound controls stay locked during setup and play, while rendering and main output volume remain local. Floppy sound settings must match. Disconnect discards the network timeline and session disk writes.

Failed startup restores the original machine. Recognized peers with incompatible builds fail immediately, and closing a browser connection drains queued diagnostics before freeing the machine. CI checks Rust/JavaScript packet-layout agreement and rejects mid-session floppy sound changes.

This connects browser peers to each other. Native/browser interoperability and TURN relays are outside this change. Update the setup/API/internals documentation and run the optimized WASM peer smoke in CI and the browser publishing workflow.

Validation:

- 3,627 native Rust unit tests and six web wrapper tests passed, including 24 netplay regressions, failed-start restoration, input routing and audio gain.
- Strict all-target/all-feature Clippy, WASM Clippy, formatting, ten page controller tests and workflow lint passed.
- Optimized A500/PAL and A1200/NTSC WASM peers confirmed matching checkpoints and final pixels under loss, reordering, duplication and asymmetric pauses, with different volume/render settings.
- Two Chromium contexts passed Host/Join, cancellation during a pending boot, restart, locked controls, initial-machine mismatch diagnostics on both peers and transmitted device-keyboard taps.
- Rebuilt native UDP peers completed 1,500 frames with identical PNG captures.
- Strict MyST site build passed using Node 24.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -477,8 +477,8 @@ jobs:
     # Guards the portability split behind the browser build (see
     # docs/guide/browser.md): the core must keep compiling without the
     # desktop frontend, and the web crate must keep compiling for
-    # wasm32-unknown-unknown. Cheap check-only job, so plain ubuntu is fine
-    # (no GPU or audio stack is reached without the `frontend` feature).
+    # wasm32-unknown-unknown, including paired rollback on a release bundle.
+    # No GPU or audio stack is reached without the `frontend` feature.
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v7
@@ -489,6 +489,9 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           save-if: ${{ github.ref == 'refs/heads/main' }}
+          workspaces: |
+            . -> target
+            crates/copperline-web -> target
       - name: Core without the desktop frontend
         run: cargo check --no-default-features --locked
       - name: Save-state fixture across optional features
@@ -507,6 +510,15 @@ jobs:
         # step never reaches it.
         working-directory: crates/copperline-web
         run: cargo fmt --check
+      - name: Browser rollback netplay on the release WASM bundle
+        working-directory: crates/copperline-web
+        run: |
+          version="$(sed -n 's/^wasm-bindgen = "=\(.*\)"$/\1/p' Cargo.toml)"
+          test -n "$version"
+          cargo install wasm-bindgen-cli --version "$version" --locked
+          RUSTFLAGS="-C target-feature=+simd128" cargo build --release --target wasm32-unknown-unknown --locked
+          wasm-bindgen --target web --out-dir pkg target/wasm32-unknown-unknown/release/copperline_web.wasm
+          node ../../tools/check-web-netplay.mjs pkg
       - name: Browser page controller tests
         working-directory: crates/copperline-web/www
         run: npm test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -505,6 +505,9 @@ jobs:
       - name: Web crate (wasm32-unknown-unknown)
         working-directory: crates/copperline-web
         run: cargo check --target wasm32-unknown-unknown --locked
+      - name: Web wrapper unit tests
+        working-directory: crates/copperline-web
+        run: cargo test --locked
       - name: Web crate formatting
         # The web crate is a standalone workspace, so the root Formatting
         # step never reaches it.
--- a/.github/workflows/wasm-demo.yml
+++ b/.github/workflows/wasm-demo.yml
@@ -9,7 +9,7 @@ name: Browser demo
 # The page shell (try/index.html) is hand-written in the website repository
 # and left alone. This workflow replaces the parts that must change together
 # with the emulator: the wasm-bindgen bundle (try/pkg), the JS glue that
-# drives the WebEmu API (try/try.js, try/render-stride.js,
+# drives the WebEmu API (try/try.js, try/netplay.js, try/render-stride.js,
 # try/serial-telnet.js, try/audio-worklet.js, sourced from
 # crates/copperline-web/www), and the AROS ROMs (try/aros).
 #
@@ -125,6 +125,7 @@ jobs:
           console.log('bundle loads:', emu.machine_summary());
           SMOKE
           node pkg-smoke.mjs
+          node ../../tools/check-web-netplay.mjs pkg
           rm pkg-smoke.mjs pkg-smoke-glue.mjs
       - name: Check out the website repository
         uses: actions/checkout@v7
@@ -143,6 +144,7 @@ jobs:
              crates/copperline-web/pkg/copperline_web_bg.wasm site/try/pkg/
           cp crates/copperline-web/www/try.js \
              crates/copperline-web/www/render-stride.js \
+             crates/copperline-web/www/netplay.js \
              crates/copperline-web/www/serial-telnet.js \
              crates/copperline-web/www/audio-worklet.js site/try/
           cp assets/aros/aros-amiga-m68k-rom.bin \
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ The manual is published at [copperline.dev/docs](https://copperline.dev/docs/) a
 - [WHDLoad Support](docs/guide/whdload.md) - Direct WHDLoad package loading
 - [Direct Executable Launching](docs/guide/run.md) - Running cross-compiled Amiga binaries
 - [Floppy Hardware Bridge](docs/guide/fluxbridge.md) - Real floppy drives via Greaseweazle
-- [Rollback Netplay](docs/guide/netplay.md) - GUI or CLI setup for direct two-player sessions with input prediction and rollback
+- [Rollback Netplay](docs/guide/netplay.md) - two-player sessions through desktop GUI/CLI or browser WebRTC, with input prediction and rollback
 - [Headless Mode](docs/guide/headless.md) - Scripted runs and screenshot/frame dumps
 - [Debugging](docs/debugger/window.md) - In-window, headless, and GDB debugging
 - [VS Code](docs/debugger/vscode.md) - Setup and illustrated source debugging; [Bartman with Copperline](docs/debugger/vscode-bartman.md) covers fork installation and visual profiling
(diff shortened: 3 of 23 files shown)


LinuxJedi (commented):
## Review by Claude (Claude Fable 5.1, via Claude Code)

Reviewer: Claude, running the pr-review-toolkit passes (code review, silent-failure hunt, test coverage, comment/doc accuracy, type design) on commit 7c36c104 against `origin/main`. Posted on behalf of @LinuxJedi.

Verification performed: `cargo check --target wasm32-unknown-unknown` clean; `cargo clippy --all-targets` clean apart from a pre-existing `src/video/nav.rs` warning; `cargo fmt --check` clean in both crates; Rust netplay lib tests 21/21; `npm test` in `crates/copperline-web/www` 8/8; `node tools/check-web-netplay.mjs pkg` passes against a local release bundle. The critical finding below was reproduced empirically against that bundle.

### Critical (1)

**Floppy-sound preferences are part of the checksummed machine state but stay live and unlocked during netplay.**

`DriveSounds { enabled, volume }` (`src/drive_sounds.rs:217`) has no serde skip and sits inside Paula (`src/chipset/paula.rs:520`), so it enters the identity hash and every 60-frame checkpoint. The wasm setters `set_floppy_sounds` / `set_floppy_sounds_volume` (`crates/copperline-web/src/lib.rs:1574`, `:1584`) have no netplay guard, `floppy-sounds` is missing from `NETPLAY_LOCKED_IDS` (`crates/copperline-web/www/try.js:117`), and the live toggle handler (`try.js:4740`) fires mid-session.

Reproduced against the bundle:

```
floppy sounds differ at boot:    netplay initial machine mismatch: use the same build, ROM, disks and deterministic machine settings
floppy sound volume differs:     netplay initial machine mismatch: ...
mid-session floppy-sound toggle: netplay desynchronized at confirmed frame 120
mono differs:                    ok (mono_output is serde-skipped)
```

The boot-time message blames ROM/disks/machine settings, which sends users hunting in the wrong place.

Fix options: mark the two preference fields `#[serde(skip)]` like `mono_output` (save-state shape change, needs a `STATE_VERSION` bump, and fixes the desktop config-level mismatch too), or gate both setters like `set_volume_percent`, add `floppy-sounds` to the lock list, and add drive sounds to the must-match list in `docs/guide/netplay.md`.

### Important (7)

1. **Non-ASCII ellipsis (U+2026) in JS string literals** at `netplay.js:246`, `:261`, `:285` and `try.js:7585`. `try.js` on main has zero non-ASCII bytes. Replace with `...`.

2. **`start_netplay` mutates the machine before the fallible call** (`crates/copperline-web/src/netplay.rs:56-67`). Volume, RTC seed, serial sink, both port devices and mouse state change before `Connection::with_transport` can fail; on failure `netplay` stays `None`, `netplay_volume` is never applied again, and the machine is left at volume 100 with a dead serial sink and both ports as joysticks. `try.js` is unaffected because `boot()` frees the machine on throw, but the embedding API in `docs/guide/browser.md` exposes it. Restore prior state in the error arm, or document that a failed call leaves the instance unusable.

3. **The 943-byte packet limit is duplicated with nothing to catch drift.** `netplay.js:5` hard-codes 943, `tools/check-web-netplay-browser.mjs:89-92` hard-codes the 111-byte header and 26-byte record, and `wire::MAX_PACKET` is `pub(super)` so nothing outside the module can read it. `netplay.test.js:88` asserts against the literal, not the Rust value, so both CI jobs pass while drifted. Export the constant (getter on `WebEmu` or a `pub const`) and assert it in `tools/check-web-netplay.mjs`.

4. **The peer that did not detect a fault only sees "Peer disconnected".** `RtcLink.close` (`netplay.js:172`) clears the incoming queue before the emulator drains the packet that would have produced the same latched error, so the second player blames the network. `tools/check-web-netplay-browser.mjs:127` works around this with `messages.some(...)`. Defer the close by one drain so both sides report the real reason, and the browser check can assert `every`.

5. **Incompatible builds stall silently for 60 s.** `wire::Packet::decode` returns `None` on a `VERSION` or `STATE_VERSION` mismatch (`src/netplay/wire.rs:65-70`) and the session times out with no hint. Pre-existing mechanism, but cached browser bundles across deploys make it likely now. Have decode report a reason and latch an explicit "incompatible build" error, or at least count rejections per reason in the timeout message.

6. **Doc claims that are wrong or overstated:**
   - `docs/guide/browser.md:216` "The guest clock starts at 2000-01-01 UTC": the browser's A500/A1200 fit no RTC (`rtc_present` defaults false), so the seed at `netplay.rs:58` acts on nothing, and `prepare_config` runs on a cloned config so its own seeding is inert.
   - `docs/guide/netplay.md:35` "keyboard joystick mode is enabled on connection": `try.js:7572` selects touch mode on touch devices.
   - `lib.rs:888` `key_raw` doc still says keys go straight to the keyboard MCU; during netplay they go to `netplay_input.set_key`.
   - `browser.md:240` says machine operations fail during netplay, but `set_port_device` (`lib.rs:1039`) and `set_floppy_speed` (`lib.rs:1596`) silently no-op. Floppy speed is guest-visible; consider `require_local_session` there.
   - `docs/internals/netplay.md:20` lists the implementing files but omits `crates/copperline-web/src/netplay.rs` and `www/netplay.js`.
   - `netplay.rs:35` "load ROM and disks before starting" is the wrong remedy once `netplay_eligible` was cleared by `run`/`load_state`; the fix is a new `WebEmu`. `netplay.rs:118` "disconnect first" names an operation the API does not have.
   - `netplay.js:85-86` "as UDP does": a UDP socket buffer tail-drops the newest datagrams; the real justification is that every packet repeats all unacknowledged inputs.

7. **Nothing in CI proves browser input reaches the shared timeline.** Both peers use the same wrapper, so a regression that drops or mis-folds `netplay_input` (`lib.rs:894`, `:982`, `:1017`) keeps both machines identical and passes every checksum and the pixel compare. Only the optional Playwright tool checks one key bit. Related gaps: no native `Connection<PacketQueue>` pair test (the two-peer path exists only behind a release wasm build); `UdpTransport::receive`'s foreign-source path is untested; `RtcLink`'s `onopen`, `onconnectionstatechange`, `ondatachannel`, `onclose` handlers never fire in `netplay.test.js`; the `take_audio` gain is never asserted; and `cargo test` for `crates/copperline-web` never runs in CI despite a 2-second native harness in `lib.rs:1621-1692`.

### Suggestions (12)

- Add `Input::set_joystick` / `Input::set_cd32_buttons` and a `BUTTONS` const. The PR adds two more copies of the undocumented bit layout (`lib.rs:986`, `:1019`); five sites now encode it, and a mismatch desyncs silently.
- Model `WebEmu` state as `Mode { Fresh, Local, Netplay(NetplaySession) }`. It removes the `unwrap()` at `netplay.rs:137` and the two fields (`netplay_volume`, `netplay_input`) that are only meaningful while netplay is active. `netplay_eligible` is a real one-way latch (reset zeroes `emulated_cck` but must not restore eligibility), so the enum expresses it better than a bool beside an `Option`.
- `UdpTransport::receive` returns `Some(0)` for foreign datagrams (`src/netplay/transport.rs:90`), relying on `decode` rejecting short input. Behaviorally identical to the old `continue`, but undocumented on the trait. Document it or return a three-state enum.
- `Options` duplicates the four `Settings` fields and `Settings::validate` runs twice per native session (`src/netplay/mod.rs:217`, `:240`). Embed `settings: Settings`; all four constructors are in-repo.
- Replace `transport_mut()` with inherent `push_incoming` / `pop_outgoing` on `Connection<PacketQueue>` so callers cannot bypass `poll`.
- Extract the RTC seed literal `946684800` shared by `mod.rs:168` and `netplay.rs:58` into a const.
- The joiner has no timeout after answering (`netplay.js:150` arms one only in `accept`). Documented as deliberate, but the joiner is left with a freed machine and locked controls; a timer or clearer status would help.
- `attach()` (`netplay.js:72`) rejects channels unless `maxRetransmits === 0`, which is not uniformly reported across engines. Only Chromium is tested; use a distinct message so a Firefox/Safari failure is diagnosable.
- `render_display_only` could itself return `Option<ContentRect>` like its sibling `render`; all four callers ignore the return, so the new `_with_content` name is unnecessary.
- `channel.onerror` discards the `RTCError` detail, and the netplay catch paths at `try.js:1193` and `:7588` log no stack. Add `console.error` before closing.
- Weak assertions in `tools/check-web-netplay.mjs`: `load_state(new Uint8Array())` at line 40 throws even without the guard (use a real `save_state()` blob); `rollbacks > 0` is only asserted for delay 0; the loss schedule depends on the 10 ms wall-clock send limiter, so a CI failure may not reproduce locally.
- CI cost: the formerly check-only job now builds `wasm-bindgen-cli` from source plus a fat-LTO release wasm with different `RUSTFLAGS` than the preceding `cargo check`. Consider `cargo-binstall` and update the job comment.

Pre-existing, not caused by this PR: `netplay_gui_peers_connect_and_can_return_to_setup_and_retry` overflows its stack under the debug profile on main as well; it passes under the CI profile.

### Strengths

- The `Connection<T: Transport>` generalization is minimal: wire protocol, rollback, timeouts and desync detection are shared verbatim and `Session` stays a type alias, so no desktop call site moved.
- `check-web-netplay.mjs` is a sharp determinism test: two real machines across A500/PAL and A1200/NTSC, loss/dup/reorder, asymmetric pauses, deliberately mismatched volume, overscan and render cadence on one peer, and a final byte-for-byte pixel compare. That test is what made the floppy-sound leak stand out as the one preference that slipped through.
- Protocol errors latch on both sides and every setup await re-checks link identity, so no path resumes as local play.
- Documentation is unusually complete, including the volume-serialization workaround, the zero-frame poll contract and the port-2 mapping rule.

### Suggested order

1. Fix the floppy-sound leak (serde-skip or gate-plus-lock).
2. Replace the four ellipsis characters, restore state on `start_netplay` failure, export the packet limit for a CI assertion.
3. Correct the doc claims and error messages above.
4. Add a native `Connection<PacketQueue>` pair test and an input-routing test in the web crate; wire `cargo test -p copperline-web` into CI.

PR #665, opened 2026-09-07
Store save states as versioned per-subsystem chunks

LinuxJedi:
Save states were a flat positional bincode image with one global version, so any change to any serialized struct invalidated every state file: `STATE_VERSION` moved nine times inside each of the last two releases, and every release discarded every user's quick-save slots.

This change makes a `.clstate` a chunk container: an uncompressed `DESC` chunk, then a zlib stream of tagged, individually versioned chunks (`CPU`, `MACH`, the two caches, and one per `Bus` subsystem: `MEM`, `CIAA`/`CIAB`, `PAUL`, `AGNS`, `COPR`, `DENI`, `BLIT`, `FLOP`, `RTC`, `GAYL`, `MOBO`, `UAEL`, `CART`, `AKIK`, `CDTV`, `ZORR`, `KEYB`, `INPT`, and a `BUS` catch-all) ending in an `END` marker. Payloads are self-describing MessagePack (structs as field maps, enum variants by name, byte vectors as `bin`), so a struct can gain a field with `#[serde(default)]`, lose or reorder fields, or widen an integer with no version change, and older states keep loading. A chunk's version moves only for a change no default can express; a `chunk::Migration` then rewrites the older payload as a value tree on load. Unknown chunks are skipped.

`Bus` keeps its single derived serde impl. `savestate/split.rs` adds a `Serializer` that streams each `Bus` field into the chunk that claims it, as a run of length-prefixed blocks written straight into the zlib encoder: a chunk opens on its first field and closes when a field of another chunk arrives, so only the catch-all `BUS` chunk (whose fields are scattered through the struct) is buffered, and anything unclaimed lands there with no table edit. The `Deserializer` counterpart pulls chunks from the stream as the derived visitor asks for the next field, decoding each `Bus` chunk straight from the stream and handing the value chunks to the machine. Neither side holds a copy of the state: a save keeps at most one 1 MiB block beyond the machine, a load at most one block beyond what it has decoded, so machines with gigabytes of memory-backed disk images keep the footprint the flat format had. In-process snapshots for reverse debugging, run-ahead and netplay rollback are unchanged positional bincode; both paths share the same restore tail. The netplay handshake carries `savestate::SCHEMA_FINGERPRINT` (crate version, container version, every chunk version) in place of `STATE_VERSION`.

Flat-format files (versions up to 80) are refused with a message saying the state must be saved again from a current build. The `END` marker must be empty and version 0 and be followed by the end of the compressed stream, in the loader and in `savestate::inspect`, so data after the marker and a corrupt zlib trailer are refused. Bus chunks decode under the decoder's nesting limit; payloads decoded from memory (descriptor, value chunks, migrated chunks) are first walked for well-formedness, exact length and depth. Payload buffers grow one MiB at a time, so a header or block claiming more than the stream holds fails without allocating it.

### Validation

- `cargo test`: 3,746 passed. New `savestate::tests` cover the chunk directory and file order, per-chunk field routing (which also proves the table names real, adjacent `Bus` fields), unknown and reordered chunks, fields a struct no longer has or did not have yet, absent required chunks, migrations and newer-chunk refusal, the end marker and trailer checks, hostile plain and block lengths and nesting, and the split/join adapters on their own. `cpu::tests::save_state_round_trip_replays_identically` exercises the file path end to end.
- `cargo clippy --all-targets --all-features --locked --offline -- -D warnings` and `cargo fmt --check` clean; `--no-default-features` native and `wasm32-unknown-unknown` checks pass.
- Headless, on the AROS factory machine and on an A1200 booting Workbench 3.1 from an HDF: save at T then screenshot at T+3 s, versus load the state then screenshot at T+3 s, produce byte-identical PNGs. A 0.19 state is refused with the flat-layout message; a 0.19 build refuses a new file with its version message.
- Fuzz seed regenerated (`fuzz/corpus/savestate/current.clstate`); the generator's stale immutable binding is fixed.
- `myst build --html --ci --strict --check-links` passed.

### Size and speed

| Measure | Flat (80) | Chunked (81) |
|---|---|---|
| AROS boot, 5 s state file | 738,543 B | 760,849 B |
| A1200 Workbench state file | 3,113,435 B | 3,184,002 B |
| Save, factory machine in memory | ~430 ms | ~3.8 ms |
| Load, factory machine in memory | ~34 ms | ~2.0 ms |

The speedup comes from feeding the zlib encoder whole blocks instead of thousands of small writes.

Documentation: `docs/internals/savestate.md` (format, block framing, chunk table, versioning rules, migration procedure), `docs/guide/ui.md`, `docs/guide/browser.md`, `docs/zorro.md`, `docs/internals/mhi.md`, `docs/internals/netplay.md`, `fuzz/README.md`, `tests/fixtures/README.md`.


LinuxJedi:
Addressed the review in d1d7935d1:

- **Load buffered the whole decompressed body** (`chunk.rs`): the `ChunkSet` is gone. `BusJoiner` now pulls chunks from the zlib stream as the derived `Bus` visitor asks for fields and decodes each `Bus` chunk straight from the stream; only the descriptor, the four small value chunks, and a payload a migration rewrites are held in memory. A load keeps at most one 1 MiB block beyond what it has decoded.
- **Save materialized every chunk before compression** (`split.rs`): `BusSplitter` now streams each chunk as a run of length-prefixed blocks into the encoder as its fields arrive, opening a chunk on its first field and closing it on the next chunk's; only the `BUS` catch-all (glue, a few MiB) is buffered. Memory-backed images of any size stream through as they did with bincode.
- **END and zlib EOF** (`chunk.rs`, `savestate.rs`): the loader and `inspect` both require the marker to be version 0 and empty, then read on and require end of stream, which drives the decoder through its trailer check. Data after the marker and a corrupt checksum are now refused (`the_end_marker_and_the_compressed_trailer_are_verified`). A stream cut off after the marker still loads, since every chunk before it arrived complete; that is documented.
- The stale bincode and wasmtime versioning notes in `Cargo.toml` and the fuzz target's module doc now describe the chunked container.

PR #667, opened 2026-09-08
Fix 68000 disassembler decode collisions and instruction lengths

vschwaberow:
- Gate CPU-dependent encodings correctly: `Bcc`/`BRA`/`BSR` `$FF` is `.B` on 68000/010 and `.L` only from 020+; full-format EA/scale  already followed the same rule.
  - Fix misdecodes that broke listings: `BKPT` vs `PEA`, `EXTB.L` vs `LEA`, `LINK.L` vs `NBCD`, illegal `MOVEA.B`, and
  immediate-to-CCR/SR limited to `ORI`/`ANDI`/`EORI`.
  - Add missing mnemonics with correct lengths (`RTD`, `TAS`, `NBCD`, `MOVE CCR,<ea>` from 010+), stop copper dumps on `$FFFF,$FFFF`,  and emit `DC.W` for a few illegal forms (memory shift on Dn/An, `ADDQ.B` to An, MOVE to PC-relative).

### DIFF
--- a/src/disasm.rs
+++ b/src/disasm.rs
@@ -260,14 +260,15 @@ fn signed_hex(v: i32) -> String {
 }
 
 /// Decode a brief-format extension word index register, e.g. `D3.W*2`.
-fn brief_index(ext: u16) -> String {
+/// Scale factors are 68020+ only; on 68000/010 the scale bits are ignored.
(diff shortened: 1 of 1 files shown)


LinuxJedi (approved):
Great catches, many thanks!

PR #672, opened 2026-09-09
Make CRT strength 0 a true sample no-op

vschwaberow:
## Summary
- Full-CRT face-edge AA darkened viewport border pixels even at strength 0 (`128` → `112`), breaking flat-grey asserts.
- Early-return the display sample when strength is zero so the documented no-op matches Scanlines/Mask.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy --lib` (only pre-existing unknown-lint noise)
- [x] `cargo test --lib 'video::window::crt_shader::' -- --test-threads=1` (23/23)

### DIFF
--- a/src/video/window/shaders/crt.wgsl
+++ b/src/video/window/shaders/crt.wgsl
@@ -139,6 +139,11 @@ fn fs_main(in: VOut) -> @location(0) vec4<f32> {
     // the strength-0 no-op.
     let wuv = mix(uv, warp(uv, u.params.w, aspect), strength);
     let base = sample_display(uv);
+    // Face AA softens the rectangle edge even when the bow is flat; skip
+    // the rest of the look so strength 0 stays a true sample no-op.
+    if (strength <= 0.0) {
+        return vec4<f32>(base.rgb, 1.0);
+    }
 
     // Scanlines follow the raster, which is straight.
     let lines = max(u.params.y, 1.0);


LinuxJedi (commented):
Review by **Claude Fable 5.1** (Claude Code `/code-review`, posted from Andrew's account). The rendering measurements below were taken on Apple Silicon / Metal, which is the CI reference GPU.

### Summary

The change is small and correct as far as it goes: at exactly `strength == 0.0` the output is now the raw sample. Three things stand out, though:

1. **The `128 -> 112` failure in the PR description does not reproduce on CI hardware.** All 23 `crt_shader` tests already pass on unpatched `main` on Metal. The 112 comes from coarse-derivative `fwidth` on the author's GPU (see the inline note on line 145), so the PR's "23/23" is the unchanged suite, and reverting the five lines still goes green.
2. **The bug is nonetheless real, but the early return only hides it at exactly 0.** `uniforms_for` (crt_shader.rs:557) produces fractional viewport heights in the real window (`ch * present_h / window_present_h`, e.g. 1016.70 px under default Smooth scaling with the status bar). With a 47.66-high viewport the unpatched shader darkens the whole bottom row at strength 0 (128 -> 105 in sRGB; 92 at frac .5). With this PR applied the same row is still 20-35% darker than the row above at strength 0.1, 0.5 and 1.0 (126 -> 103 / 119 -> 90). So the knob now jumps from a clean edge to a dark hairline on its first step. The root cause is the AA band being centred on the face edge; see the inline note on line 144 for a one-line fix that makes the guard unnecessary.
3. **No test pins the fix.** See below.

### Findings in files outside the diff (cannot be attached inline)

**`src/video/window/app_menus.rs:529` -- the Softer step never reaches exact 0 from most starting points.** `self.shader_strength + step * dir` is plain f32 arithmetic with no snapping. From 0.5 (or 0.4, 0.3, 0.6) the press that makes the OSD read "Shader strength: 0%" lands on `1.4901161e-08`, not `0.0`. The menu (`src/video/menu.rs:734`, `available(s.shader_strength > 0.0)`) still offers Softer, but the shader's new `strength <= 0.0` guard is *not* taken, so the face-AA darkening this PR exists to remove is still rendered at a displayed 0%. One more press reaches exact 0 and the line disappears: two states both labelled 0% render differently. In a player build `persist_player_prefs` (window.rs:5692) writes the 1.49e-8 verbatim and `validate.rs:380` accepts it, so the "0%" darkening survives restarts. Suggested fix: snap the stepper to a multiple of `SHADER_STRENGTH_STEP` (e.g. `((v / step).round() * step).clamp(0.0, 1.0)`), or treat near-zero as zero before the uniform upload.

**`src/video/window/crt_shader.rs:1262` -- no regression test can catch a revert.** `render_preset` always renders an integer 64x48 / 128x96 viewport, where every edge fragment sits 0.5 px inside the face and the AA band yields `face == 1`, so `presets_render_the_display_region_only` and `a_magnified_display_never_blends_in_the_status_bar` pass on unpatched `main`. A later edit that hoists the face multiply above the early return, or a comparison that misses exact 0, passes `cargo test` on every CI leg. A test that clones `render_preset` with a fractional viewport height (override `viewport.3` and `uniforms.size[1]` to e.g. 47.66, as `uniforms_for` does in the real window) and runs `assert_flat_grey` on `Crt` at strength 0 fails on `main` (last row 105 vs 128, tol 2) and passes on this PR, deterministically on Metal. Keep it to strength 0: at 0.1 the PR still renders that row ~20% darker, which is the separate line-144 finding.

**`docs/internals/video.md:913-916` -- the documented mechanism is now wrong.** The text still says the no-op holds "through the warped coordinate, so strength 0 has no off-face region at all and the no-op invariant holds". This PR's own rationale shows geometry alone does not deliver it; the invariant now rests on an early return no doc mentions. AGENTS.md asks for docs to move in the same change, and the previous two face-edge commits (9f92e4281, 96fa24b29) updated video.md accordingly. Only the mechanism sentence needs correcting; line 956, `crt_shader.rs:22-23` and `configuration.md` describe the outcome, which this PR makes literally true.

### Checked and dropped

NaN strength (rejected by both config and env paths); gating the pass on the Rust side (deliberately not done, and it would flip the bezel compositing path); `select()` on `face` instead of an early return (pixel-identical, and naga treats uniform-buffer-derived conditions as uniform, so the early return is not a validation hazard); the web `try.js` GLSL port (hardcodes strength 1.0, so no visible bug, only a now-stale "keep in step" comment); a bezel-join seam (the bezel discards and blends toward black, so it neither masks nor amplifies the bottom-row defect).

src/video/window/shaders/crt.wgsl
```diff
@@ -139,6 +139,11 @@ fn fs_main(in: VOut) -> @location(0) vec4<f32> {
     // the strength-0 no-op.
     let wuv = mix(uv, warp(uv, u.params.w, aspect), strength);
     let base = sample_display(uv);
+    // Face AA softens the rectangle edge even when the bow is flat; skip
+    // the rest of the look so strength 0 stays a true sample no-op.
+    if (strength <= 0.0) {
```
LinuxJedi, line 144: **Root cause is untreated; the guard only hides it at exactly 0.** The one-pixel AA band at line 198 is centred on the face edge (`face = 1.0 - clamp(d / aa + 0.5, 0.0, 1.0)`), so its inner half lies *inside* the viewport and dims any fragment whose centre is within half a pixel of a straight edge. Under the real window's fractional viewport heights (`uniforms_for`, crt_shader.rs:557) that is the whole bottom row at every strength > 0. Measured on Metal, sRGB target: strength 0.1 renders the bottom row 126 -> 103 at frac .66 and 119 -> 90 at frac .5, and the same 1-px dark line persists at 0.5 and 1.0 where the design says the mid-edge is flush. Because `rc` cancels on a straight edge and the bow is flush at the mid-edges, the strength-0.1 row equals the unpatched strength-0 row.

Deeper fix: keep the fade band outside the face,

```wgsl
let face = 1.0 - clamp(d / aa, 0.0, 1.0);
```

which guarantees `face == 1` for every rasterised fragment when `rc == 0` on any GPU, removes the need for this early return, and still passes `the_crt_face_edge_is_hard_at_partial_strength` and `the_crt_preset_darkens_the_corners`.

src/video/window/shaders/crt.wgsl
```diff
@@ -139,6 +139,11 @@ fn fs_main(in: VOut) -> @location(0) vec4<f32> {
...
+        return vec4<f32>(base.rgb, 1.0);
```
LinuxJedi, line 145: **Why CI saw 128 and you saw 112: `fwidth` at the viewport boundary is implementation-defined.** WGSL allows coarse or fine derivatives, and the line-132 uv clamp pins helper lanes to `d = 0`. Integer 64x48 viewport, strength 0, unpatched: the bottom-right corner quad has d = (-1.5, -0.5; -0.5, -0.5) px. Coarse derivatives give `fwidth = 2` px so three lanes get `face = 0.75` = sRGB 112; fine derivatives give 1 px and 128. That is exactly the split between the CI reference GPU (Metal, fine) and your GPU (coarse), and it flips with the parity of the display-rect origin. With this PR that driver dependence persists at every strength > 0; the strength-0 assert merely stops exercising it.

Because `d` is isotropic in physical pixels by construction (`half = (1, aspect)` makes one pixel exactly `2 / u.size.x` on both axes; the warp Jacobian scales it by at most 1.14 at full strength), an analytic band

```wgsl
let aa = 2.0 / max(u.size.x, 1.0);
```

gives a 0.88-1.0 px edge identical on every backend and makes `presets_render_the_display_region_only` GPU-independent. This complements the strength-0 fix rather than replacing it: with a fractional viewport the last row sits at d = -0.16 px and any centred band softens it.

src/video/window/shaders/crt.wgsl
```diff
@@ -139,6 +139,11 @@ fn fs_main(in: VOut) -> @location(0) vec4<f32> {
     // the strength-0 no-op.
     let wuv = mix(uv, warp(uv, u.params.w, aspect), strength);
     let base = sample_display(uv);
+    // Face AA softens the rectangle edge even when the bow is flat; skip
+    // the rest of the look so strength 0 stays a true sample no-op.
```
LinuxJedi, line 143: **Four comments still credit the no-op to the fades and the flat bow, which this guard makes false.** Lines 138-139 (`wuv` mix "keeping the strength-0 no-op"), 170-177 ("at strength 0 the bow is flat, nothing lies off the face, and the no-op invariant holds"), 185-187 (radius fade "keeping the strength-0 no-op") and 199-202 (glow "so strength 0 stays a no-op"). After this patch that code never runs at strength 0, and the geometric claim is false under a fractional viewport. Meanwhile the fades' real reasons (continuity at 0+, and bezel mirroring via `bezel.rs:99/282` `FACE_CORNER_RADIUS * strength` and the coordinate mix at 130-134) are stated nowhere, and this new comment does not say that `face` is the only unscaled term.

The risk is a CI-silent regression in either direction: a maintainer trusting 170-177 deletes this guard as redundant (23 tests still pass on the integer viewport, the real window's strength-0 bottom row darkens again), or a maintainer trusting this guard deletes the "redundant" radius/coordinate/glow fades (tests still pass, but the bezel aperture disagrees with the face at every partial strength and the picture gains a `GLASS_GLOW` step and a full-radius corner crop at 0+). Suggest rewriting 138-139, 170-177, 185-187 and 199-202 to say the guard owns the strength-0 identity, the fades exist for continuity and bezel mirroring, and `face` is the sole unfaded term.

src/video/window/shaders/crt.wgsl
```diff
@@ -139,6 +139,11 @@ fn fs_main(in: VOut) -> @location(0) vec4<f32> {
     // the strength-0 no-op.
     let wuv = mix(uv, warp(uv, u.params.w, aspect), strength);
     let base = sample_display(uv);
```
LinuxJedi, line 141: **Minor: hoist the guard above `aspect` and `wuv`.** `base` depends only on `uv`, and neither `aspect` nor `wuv` is used on the return path, so placing the `if` directly after `let base = sample_display(uv);` and moving lines 135 and 140 below it is semantically identical (pure arithmetic; `textureSample` stays ahead of the uniform branch) and matches the prologue order in `scanlines.wgsl`, `mask.wgsl` and the `configuration.md` template (uv, base, strength, then the look). As written, the comment at 136-139 still reads as if the warp geometry secures the no-op, which is the claim this PR says is false, and a reader has to confirm `base` was sampled from `uv` rather than `wuv` before trusting the early return. Readability only; the dead-warp ALU at strength 0 is negligible.

PR #673, opened 2026-09-09
Ship copperline-ctl and copperline-import-uae in every release package

LinuxJedi:
A user reported that `copperline-ctl` is missing from the Windows build. It was wider than Windows: every pre-built package except Homebrew shipped only the `copperline` executable.

## Why

`copperline-ctl` and `copperline-import-uae` are default-feature binaries, so every packaging job already compiled them. The staging steps then copied the emulator alone. Homebrew got them only because `cargo install` installs every default binary. The DAP guide and the VS Code extension tell users to have `copperline-ctl` on PATH, so dmg, zip, AppImage and Flatpak users could not use the debug adapter or the MCP server without a source build, and no document said so.

## Changes

- **Windows zip:** `build-zip.ps1` copies both `.exe` files beside `copperline.exe`. The zip smoke job fails if either is missing or its `--help` exits non-zero. Bundled README gains a "Command-line tools" section.
- **macOS dmg:** the build job uploads all three binaries as its per-architecture slice. `build-dmg.sh` lipo-joins each into `Contents/MacOS`, signs the two tools explicitly before the bundle-level `codesign --deep` (which does not reliably reach extra executables in `Contents/MacOS`), and the smoke job checks both are universal and start natively and under Rosetta. README updated.
- **AppImage:** a single-entry-point image cannot expose a second binary, so the script also produces `Copperline-X.Y.Z-<arch>-tools.tar.gz` holding both tools plus the license. It is uploaded as a workflow artifact, attached to releases alongside the net-helper tarball, and smoke-tested.
- **Flatpak:** both tools are installed into `/app/bin`, reachable via `flatpak run --command=copperline-ctl dev.copperline.Copperline`.
- **`--help` exit status:** both tools returned 1 for `--help`, which the new smoke checks rely on. `-h`/`--help` now print usage to stdout and exit 0; bad arguments still exit 1.
- **Docs:** getting-started gains a "Command-line tools" table listing where each package puts the tools. The DAP chapter, VS Code guide, UAE import guide and the extension README point at it instead of implying a source build is required.

## Validation

- Both binaries rebuilt in release: `--help` exits 0, bad arguments exit 1. `cargo fmt --check` and `cargo clippy` clean on the two binaries.
- Local dry run of `build-dmg.sh` with `PREBUILT_DIR` against host-architecture binaries: the bundle passes `codesign --verify --deep --strict`, both tools are present, signed and run from the mounted image, and the emulator boots the bundled AROS ROM from the read-only volume.
- Shell scripts pass `bash -n`; all four YAML files parse; `actionlint` reports only the pre-existing info notes on `ls` usage.
- The Windows and AppImage paths get their first real run in CI here (no `pwsh` locally, and the tools tarball step only runs on Linux).

## Size note

`copperline-ctl` links the whole emulator library, so it adds about 52 MB per architecture uncompressed to the dmg and zip before compression. The release profile's symbol handling is left alone, since stripping would also affect crash-report backtraces from the main binary.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -56,6 +56,7 @@ jobs:
         run: |
           echo "path=$(ls Copperline-*.AppImage | head -n1)" >> "$GITHUB_OUTPUT"
           echo "helper=$(ls Copperline-*-net-helper.tar.gz | head -n1)" >> "$GITHUB_OUTPUT"
+          echo "tools=$(ls Copperline-*-tools.tar.gz | head -n1)" >> "$GITHUB_OUTPUT"
       - name: Smoke test the AppImage
         # Boot the finished AppImage from an empty directory: the run must
         # find the bundled AROS ROM through the AppDir's
@@ -74,6 +75,12 @@ jobs:
           cd "$smoke"
           "$appimage" --noaudio --screenshot-after 10 smoke.png
           test -s smoke.png
+          # The command-line tools tarball is a release asset of its own;
+          # both tools must be in it and must start.
+          tar -xzf "$GITHUB_WORKSPACE/${{ steps.artifact.outputs.tools }}"
+          tools="$(ls -d Copperline-*-tools)"
+          "$tools/copperline-ctl" --help >/dev/null
+          "$tools/copperline-import-uae" --help >/dev/null
       - uses: actions/upload-artifact@v7
         with:
           name: copperline-appimage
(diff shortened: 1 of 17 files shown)


LinuxJedi:
Addressed in the follow-up commit:

1. `-h`/`--help` is now recognised anywhere on the command line in both tools (`copperline-ctl --mcp --help`, `copperline-ctl status --help`, `copperline-import-uae --from winuae --help` all print usage and exit 0). No mode, subcommand or option takes those tokens as a value, so position does not matter. The per-parser help arms this made unreachable are removed.
2. The extension README now lists the Flatpak `/app/bin` location with the other packages.
3. The AppImage tools tarball carries a `README.txt` (`packaging/linux/copperline-tools-README.txt`) describing both tools and the `COPPERLINE_BIN` lookup order, with a link to the getting-started section.

Verified: both binaries rebuilt in release, the help matrix above exits 0 and bad arguments still exit 1; `cargo fmt --check` and `cargo clippy` clean; `bash -n` clean.

PR #678, opened 2026-09-10
Show what each inspector tab holds, and fix two Debug workspace bugs

LinuxJedi:
Three related changes to the Debug workspace, coming out of a discussion about
why the "Close inspector" button read as pointless.

## The tab strip did not show that "open" was a state

An inspector open behind the current one was drawn exactly like one that had
never been opened. With no way to see that an inspector was open, closing one
showed no visible change -- which is what made the strip-level **Close
inspector** button look like it did nothing. It also sat at the far end of the
strip from the tab it acted on, and could only ever close the *selected*
inspector.

Each tab now shows which of its three states it is in, and carries its own
close box:

| Tab | Meaning |
|---|---|
| Dimmed, no dot | Not open. Click to open it. |
| Filled, with a dot and a close box | Open. Its state and its capture are there whether or not it is on screen. |
| Blue | Open and on screen. |

The close box closes the inspector it is drawn on, background ones included,
which the old button could not reach at all. The dot reports whether that
inspector's capture is armed on the machine in front of you -- the only
indication of the cost an open inspector keeps paying once you are back in
Play.

**Return to Play** becomes a **Play / Debug** switch. Leaving Debug is a change
of view, not a dismissal: the inspectors stay open and keep capturing, and a
two-state switch reads that way where a button did not. The tabs *within* an
inspector are underlined rather than blue-filled, so the two levels of the
header stop looking alike.

## Bug: inspectors went dead when the machine was replaced

Reported symptom: with the debugger open, return to Play, change the machine
configuration, reopen the debugger -- the debugger runs but the Frame Analyzer
will not until a Pause and a Run.

Everything an inspector captures with is armed on the bus and the CPU, while
the panels are host state that outlives the machine. A machine built by the
launcher's Run comes up with none of that arming, and the still-open panels
skip the arming they only perform on first open. Three more casualties of the
same cause, from the same reproduction: Recent PCs stayed empty, the reverse
controls (`< Step` / `< Run`) were dead, and the Memory tab's heat map recorded
nothing. `run_machine` now re-arms whatever is still open, under the same
ownership rule the Memory tab applies.

## Bug: the beam diagram resized on every frame

Reported symptom: on an interlaced screen the diagram grows and shrinks
constantly and everything below it jumps; a pointer held on one row reads 312
and 313 alternately.

An interlaced signal alternates a long field and a short field one line shorter
(PAL 313/312, NTSC 263/262). The capture reports the field it actually took, so
laying the diagram out against it resized the pane twice a frame. Size, pick and
overlays now go against the long field, with the row clamped to the field
actually captured. A short field has no last line, so a position over it reads
the line before -- that one is real data, not layout.

## Tests

- `running_a_new_machine_rearms_the_open_inspectors` -- verified to fail
  without the re-arm (analyzer capture, heat map, reverse ring, Recent PCs).
- `alternating_interlace_fields_do_not_resize_the_beam_diagram` -- verified to
  fail without the fix (`[574x395.7], [574x394.5], [574x395.7]`).
- `a_background_tabs_close_box_closes_that_inspector`,
  `a_closed_tab_has_no_close_box_and_opens_on_click`,
  `the_tab_strip_reports_open_and_capturing_per_inspector`,
  `the_mode_switch_returns_to_play_without_closing_anything`.

3803 unit tests pass; `cargo clippy --all-targets` and `cargo fmt --check`
clean. Docs and the seven `-egui` screenshots are regenerated in the same
change.

Not covered: the chrome is verified through the offscreen renderer and
unit-level click simulation, not a live window session, so hover states
(close-box highlight, tooltips) are worth an eyeball.


LinuxJedi:
Thanks — all three addressed in 27b11c5.

**Programmable frame totals misclassified as short fields.** Correct, and the fix
was already sitting in the chipset: `Agnus::nominal_frame_lines()` returns the
long field for an interlaced signal *and* a programmable VARBEAMEN total as it
stands. The capture now carries that as `FrameBusTrace::nominal_rows` and the
analyzer lays out and picks against it, so the parity guess is gone. A 200-line
programmable frame is laid out as 200 with no bottom-row clamp.

Covered by `the_capture_carries_the_frame_height_agnus_reckons` (interlaced
fields alternate while the frame height holds still; an ECS VARBEAMEN 200-line
frame reports 200/200) and an assertion in
`alternating_interlace_fields_do_not_resize_the_beam_diagram`. Verified the
latter fails if the parity rule is put back.

**Widget metadata.** Correct that the replacements described themselves less
than the `selectable_label`s they replaced. Tabs, sub-tabs and mode segments now
report label, role and selected state; the close box is painted rather than
typed, so it carries an explicit `Close <inspector>` label. Worth noting for
anyone reading this later: `egui-winit` is built here with
`default-features = false` and no `accesskit` feature, so nothing consumes this
yet — it was a latent regression rather than a live one, but the controls
should not describe themselves less than what they replaced.

**Contrast.** Confirmed at 2.74:1 — a closed tab is dimmed but still an
actionable control, so it needs to stay legible. `MUTED` is now 98,105,117 =
4.84:1 on the panel background, still clearly apart from the ink an open tab
uses (14.3:1), so the state distinction survives.

PR #679, opened 2026-09-11
Band-limit Paula's output onto the host mixer grid

LinuxJedi:
Reported as a ringing, clipping 10 kHz tone in Amiga Test Kit's audio page,
where FS-UAE sounds like real hardware. Two independent faults in the path
from Paula to the host mixer.

## No anti-aliasing on the decimation

Paula's output is a staircase stepping on the 3.546895 MHz colour-clock grid,
and `advance_audio` read it once per 44.1 kHz frame, so everything above the
mixer's Nyquist folded back into the audible band. Test Kit's 10 kHz square
(two samples at AUDxPER 177, full volume on all four channels) has harmonics
at 30, 50 and 70 kHz; measured on A500/KS1.3 with the filter off:

| freq | level | what it is |
|---|---|---|
| 10019.5 | 0 dBc | the tone |
| 14041.5 | **-10.0 dBc** | 3rd harmonic folded at 44100-30058 |
| 5997.5 | **-15.3 dBc** | 5th folded |
| 18063.8 | -17.9 dBc | 7th folded |
| 1975.2 | -19.8 dBc | 9th folded |

plus +/-71.5 Hz sideband pairs from the beat between Paula's 20038.95 Hz
sample rate and 44.1 kHz. An independent model of the exact signal confirmed
point sampling reproduces that spectrum line for line, that integrating over
the output interval (what UAE calls "anti") accounts for the FS-UAE
comparison, and that band-limited decimation leaves a single clean line.

The mix now runs at 4x the mixer rate with exact-area integration, then a 4:1
windowed-sinc decimation per channel. Worst in-band image goes from **-10 dBc
to -51 dBc**. Four was chosen by measurement: 2x leaves a -28 dBc spur, and
neither 8x nor a longer kernel improves on 4x.

**The LED filter made it worse, not better.** Real hardware filters before the
sampling; we filtered after, so the 4 kHz filter attenuated the genuine tone
and left the aliases that had already folded below its cutoff. The 1975 Hz
alias ended up 5.8 dB *above* the tone -- a real A500 goes nearly silent on
this test, we produced a 2 kHz buzz. The tone is now the loudest thing in the
output with everything else at least 43 dB down.

Paula's own zero-order-hold images are deliberately kept: they are real output
ahead of the hardware's analogue filter, so the 500 Hz test still carries its
images at 19.5 and 20.5 kHz.

## No headroom

Two channels reach each side and each was scaled by `1/(128*64)`, so full
scale on all four gave +/-1.98 with nothing clamping between there and the
host device. **97.7% of the 10 kHz capture sat outside +/-1.0.** Full scale is
now both channels saturated plus the overshoot a band-limited waveform needs,
taken from the decimation kernel: the sum of its absolute taps, 1.5792, is the
most it can make of input bounded by full scale, whatever that input is.

Deriving it instead from the shortest period a channel can be given would not
hold. `PAL_AUDIO_MIN_PERIOD_CCK` / `NTSC_AUDIO_MIN_PERIOD_CCK` describe what
audio DMA can sustain, are defined but never referenced, `aud_percntrld` takes
`AUDxPER` as given, and a CPU feeding `AUDxDAT` in IRQ mode is held to
neither. Output is 10.0 dB quieter than the old scaling.

## Results

| | before | after |
|---|---|---|
| worst in-band spur | -10.0 dBc | **-51.3 dBc** |
| 10 kHz test peak | 1.98 (97.7% of samples clipped) | 0.80 |
| LED filter on | alias 5.8 dB **above** the tone | tone loudest, rest >=43 dB down |
| Test Kit module peak | 2.13 (clipping) | 0.67 |
| 500 Hz test peak | 1.98 (65% clipped) | 0.63 |

("before" peaks above are the measured after-values scaled by the exact ratio
of the old and new divisors, 3.16; the mix is linear in it.)

## Verification

- 3834 unit tests pass; `clippy` and `fmt --check` clean.
- Five new Paula regression tests (folded-harmonic levels via DFT; a sweep
  over duty cycles and periods -- down past what DMA can sustain -- asserting
  nothing Paula can play leaves +/-1.0; DC unity through the chain; the
  headroom constant against the kernel's actual peak gain; and the mixer
  cadence surviving a resume at each oversample slice) and three `Decimator`
  tests.
- Six existing tests updated: four for the new scale, two because the
  decimator's ~0.2 ms group delay means the first frame is no longer the
  settled value.
- **Emulated timeline untouched.** Slice counts are byte-identical before and
  after (6305778 on boot, 21777906 on a 45 s four-channel run) and the golden
  probe renders are unchanged. No `PAUL` chunk version bump is needed:
  `host_sample_acc` keeps its meaning, and the new fields are
  `serde(default)`.
- Cost: ~2% slower with all four channels at AUDxPER 177, ~2% on boot.
- `docs/internals/audio.md` documents the decimation, the level reference and
  why the LED filter can only sit where it does once decimation is
  band-limited.

## Known gaps, deliberately not addressed here

- The mix bus still has no headroom for CD-DA plus Paula. CD-DA arrives at
  +/-1.0 and is line-mixed after Paula's scaling, so a CD32 playing a
  full-scale track under a full-scale module can reach ~1.7 (it was ~3.1).
  Fixing that means deciding the hardware line-level balance between
  Akiko/CDTV CD audio and Paula, which wants a measurement first. Same for
  MT-32, Toccata and MHI.
- No always-on output RC network is modelled for any machine, only the
  switchable LED filter, so A500 and A1200 still measure identically. Real
  machines differ.
- Output volume caps at 100% and everything is now 10 dB quieter, so the
  status-bar slider can only attenuate.

### DIFF
--- a/docs/internals/audio.md
+++ b/docs/internals/audio.md
@@ -6,6 +6,82 @@ routes the master mix and individual sources to playback or WAV capture.
 Mixing, LED filtering, volume, and stereo width are applied in
 `Paula::push_mixed_frame` (`src/chipset/paula.rs`).
 
+## Getting Paula onto the host grid
+
+Paula's four channels are a staircase: each holds its 8-bit sample for
+`AUDxPER` colour clocks, so a channel steps at up to ~28.6 kHz and the
+summed output only ever changes on the 3.546895 MHz colour-clock grid.
+Landing that on the mixer's 44.1 kHz grid is a decimation by ~80.4, and
+doing it by reading the staircase at each output instant would fold
+everything above the mixer's Nyquist straight back into the audible band.
+The staircase is rich up there: a full-scale 10 kHz square (two samples at
+`AUDxPER` 177, what Amiga Test Kit's audio page plays) carries harmonics at
+30, 50 and 70 kHz, which point sampling turns into tones at 14.0, 6.0 and
+18.1 kHz -- inharmonic, so heard as ringing rather than brightness, and
+loud, the worst of them only 10 dB under the tone itself.
+
+So the mix runs in two stages, both in `Paula::advance_audio`:
+
+1. **Exact integration to an oversample grid.** `PAULA_OVERSAMPLE` (4) times
+   the mixer rate, so 176.4 kHz. `advance_audio` splits its span so that no
+   oversample instant and no channel period expiry falls strictly inside a
+   step (`cck_until_mix_input_changes`), which means each channel's
+   contribution is constant across the step and `channel_area` can
+   accumulate `sample * volume * clocks` exactly. Dividing by the interval
+   width at each oversample instant is then a true box average of the
+   staircase, not a sample of whichever step the instant landed on.
+2. **Windowed-sinc decimation to the mixer grid.** One
+   `audio::resample::Decimator` per channel, 4:1, sharing the 64-tap
+   Blackman-windowed kernel `Resampler` uses. Per channel rather than per
+   side because decimation is linear -- filtering each channel and summing
+   is the same signal as filtering the sum -- which keeps the per-channel
+   stem taps exactly consistent with the mix they add up to.
+
+Neither stage touches the emulated timeline: `advance_audio` runs the same
+channel state machine over the same colour clocks either way, and the extra
+splitting only changes how often it is re-entered. The box stage needs the
+oversample rate well above Paula's own; the sinc stage needs it low enough
+that the kernel's transition band stays clear of the top of the audio band.
+Four satisfies both, and puts the worst surviving image about 47 dB under
+the tone. Raising the oversample rate or lengthening the kernel does not
+measurably improve on it.
+
+The genuine zero-order-hold images of Paula's own sample rate are *not*
+filtered out, because they are real output: the 500 Hz sine test at
+`AUDxPER` 177 keeps its images at 19.5 and 20.5 kHz, exactly as the
+hardware produces them ahead of its analogue filter.
+
+### Level
+
+One channel's DAC level is `sample * volume`, -8192..8128. Two channels
+reach each side, so full scale is both of them saturated at full volume --
+and then the band-limited waveform overshoots the steps it came from, since
+taking the harmonics away leaves something taller than the square that
+carried them. The most any Paula channel can overshoot is set by its
+shortest legal period: `AUDxPER` 124 holds a sample for a little over six
+oversample intervals, and the most a run that long can draw out of the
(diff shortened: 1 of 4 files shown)


LinuxJedi:
All three addressed in dfdf0cd1. Both findings were right, and the first one
was righter than it looked.

**Minimum period (2 findings).** Correct that 124 is the NTSC figure and
`PAL_AUDIO_MIN_PERIOD_CCK` is 123. But recomputing the bound from 123 would
still have been wrong, because neither constant is a floor on anything: both
are defined and *never referenced* anywhere in the tree, `aud_percntrld` takes
`AUDxPER` as given, and a CPU feeding `AUDxDAT` in IRQ mode is not held to the
period audio DMA can sustain. So the whole shape of the argument was unsound,
not just the number in it.

The headroom now comes from the decimation kernel: the sum of its absolute
taps is the most it can make of input bounded by full scale, whatever the
input is. That is 1.5792 against the 1.3241 the period argument gave, costing
a further 1.5 dB (10.0 dB total below the old scaling, up from 8.5 dB) and
buying a guarantee that holds unconditionally.
`the_mix_headroom_covers_the_decimation_kernel` now asserts the constant
against `Decimator::peak_gain()` so the two cannot drift apart, and the
no-clipping sweep runs down past what DMA can sustain (`AUDxPER` 8, 20, 41,
80, 113, 123 ...). The docs describe the two limits for what they are.

**Legacy save-state phase.** Confirmed, and the invariant you identified
(`count == oversample_index()`) was exactly the problem: it made the phase
redundant state, agreeing with the accumulator during a run and silently
diverging on a restore that started it at zero part way into a frame.

Rather than resynchronise the counter on load, I removed it. `Decimator` now
holds tap history only -- `push` records a sample, `output` reads the
band-limited value, and `advance_audio` decides when it wants one, from the
accumulator. The mixer phase lives in exactly one place, so there is nothing
left to contradict it and no load-time fix-up to forget. It also dropped
`count` from the serialized form.

`the_mixer_cadence_follows_the_accumulator_not_the_decimators` covers it:
resuming at each oversample slice in turn, with decimators that have never
seen a sample, the mixer frame in progress must still be finished. Under the
counting version that failed for slices 1-3, which is the reported bug.

No change to the measured result: worst in-band image still -51 dBc, slice
counts identical, and the rework made it about 2% slower than before the
band-limiting rather than 4%, since `push` no longer branches and the kernel
is applied once per mixer frame instead of once per channel per slice.
3834 tests pass; clippy and fmt clean.

PR #712, opened 2026-09-18
launcher: the Netplay row says Enabled or Disabled like the rest

hobbo91:
## The issue

The Netplay page opened with the launcher's last On/Off button. Every other switch is a `< Enabled >` / `< Disabled >` stepper since the launcher sweep (#486), and the one button left over looked out of place at the top of its page.

## The fix

- **The Netplay row is a stepper** on both forms of the page (Direct IP and Internet); either arrow flips it.
- **What flipping it does is unchanged.** Enabling still settles the machine on the other pages (serial off, interpreter, power on, digital ports), and the row is still the one control on the page that is live while netplay is off.
- The two special cases that existed only for the button are removed.

Purely cosmetic: no config, CLI or save-state change. The netplay guide's wording ("Enable **Netplay**") already fits; its two launcher pictures are regenerated.

Netplay page

## Testing

- The tests that pressed the button now press the stepper; a new test, `the_netplay_row_is_an_enabled_disabled_stepper_on_both_pages`, pins the row kind, the label, and that enabling still switches serial off.
- `cargo test --locked`: 3,876 passed, 43 ignored.
- `cargo fmt --check`, `cargo clippy --all-targets --all-features --locked -- -D warnings`, and the `--no-default-features` and `frontend,cpu-jit` checks are clean.
- Checked on macOS only; the change is in the pixel-drawn launcher, which is the same code on every host.


LinuxJedi (approved):
Great catch! Many thanks!

PR #714, opened 2026-09-18
Document the copperhf product ID and drop a duplicated clause

sidick:
Two documentation-only fixes found while surveying the board plumbing.

- `docs/zorro.md`: the Copperline manufacturer ID product table listed products 1-7 but omitted product 8, the copperhf virtual block-storage board (`PRODUCT_COPPERHF` in `src/zorro.rs`). Added the row in the style of the existing ones.
- `docs/internals/architecture.md`: the determinism bullet about live host I/O repeated its own final clause ("or services. Physical floppy drives also require wall-clock pacing.") -- a leftover from an earlier edit. Removed the stray line.

No code changes. `cargo fmt --check` is clean; clippy does not read these files. The MyST docs build was not run (`myst` is not installed on this machine) -- both edits are plain table/prose lines with no directive or cross-reference syntax.


LinuxJedi (approved):
Many thanks!

